PR #152, opened 2021-09-22
Modify the existing shell commands to become more sudo friendly

Pouriya-Inmanta:
# Description

Modifies the existing shell commands to become more `sudo` friendly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,4 @@
-# v 1.3.0 (?)
-Changes in this release:
-
-# v 1.2.2 (2021-08-22)
+# v 1.2.2 (?)
 Changes in this release:
 - Modify shell commands to be more sudo friendly.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
-# # v 1.3.1 (2021-08-22)
-Changes in this release:
-- Modify shell commands to be more sudo friendly.
-
 # v 1.3.0 (?)
 Changes in this release:
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# # v 1.3.1 (2021-08-22)
```
wouterdb, line 1: The changelog entry without date is unreleased, so you can add it there (or make it v1.2.2)

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # v 1.3.0 (?)
 Changes in this release:
 
+# v 1.2.2 (2021-08-22)
```
wouterdb, line 4:
```suggestion
# v 1.2.2 (?)
```

@sanderr should any other file be updated to build release 1.2.2?

wouterdb (changes requested):

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-# v 1.3.0 (?)
+# v 1.2.2 (?)
```
sanderr, line 1: So, as discussed on slack, I would undo this change. Keep the change entry, but revert the change to the version here.

sanderr (changes requested):

CHANGELOG.md
wouterdb: oeps, sorry, bad advice from my side.

wouterdb (approved):

sanderr (approved):

PR #182, opened 2022-02-07
Issue/178 project install

### LINKED ISSUE #178: Add support for inmanta~=5.0
With iso5, the module dependencies are not installed automatically on each compile.  They are installed when doing an `update and recompile` (recompile with force_update set to True) or when using the command `inmanta project install` (in the venv of the environment's compiler).

This is currently not supported by pytest-inmanta-lsm.

What this ticket should implement:
- The logic to get the server version: `api/v1/serverstatus`
- Check if we have a server with iso5 or later
- If previous condition pass, install modules dependencies via ssh
  ```
        # Run inmanta project install
        LOGGER.debug("Install project dependencies")
        subprocess.check_output(
            SSH_CMD
            + [
                f"-p {self._ssh_port}",
                f"{self._ssh_user}@{self.host}",
                f"sudo sh -c 'cd {server_path}; source .env/bin/activate; PIP_INDEX_URL=https://artifacts.internal.inmanta.com/inmanta/dev inmanta project install'",
            ],
            stderr=subprocess.PIPE,
        )
  ```

sanderr:
# Description

The `remote_orchestrator` fixture rsyncs a project to the server. Because of this unconventional approach an iso5 server does not pick up on the fact that a project installation should be performed (it has no way to make this decision). As a result, the fixture is broken for modules with Python dependencies (unless they happen to be dependencies of the inmanta product as well). This PR adds a test to reproduce the issue and fixes it by running a project installation on each project sync.

closes #178

# Self Check:

- [x] Changelog entry
- [x] ~~End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )~~

# Reviewer Checklist:


wouterdb (approved):

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -231,6 +233,34 @@ def sync_project(self) -> None:
...
+            raise Exception(f"Failed to get server status for {self._host}")
+        server_version: Version
+        try:
+            server_version = Version(server_status.result["data"]["version"])
+        except (KeyError, TypeError):
+            raise Exception("Unexpected response for server status API call")
+        # iso5 requires explicit project installation
+        if server_version >= Version("5.dev"):
+            venv_path: str = os.path.join(server_path, ".env")
+            # venv might not exist yet so can't just access its `inmanta` executable -> install via Python script instead
```
edvgui, line 247: If the venv is not created, will this just work? Isn't it necessary to call `.use_virtual_env` on the `VirtualEnv` object created by the project?  I am genuinely asking, I failed to verify this.  If you know it is the case, you can simply close this, I do not think that a test case is necessary for this.

examples/test_service/tests/test_deployment_failure.py
```diff
@@ -11,7 +11,8 @@
 SERVICE_NAME = "test-service"
 
 
-def test_full_cycle(project, remote_orchestrator):
+@pytest.mark.parametrize("fail", (True, False))
```
edvgui, line 14: I am failing to see the relation between the extension of this test case and your change, is there any?  Or is it simply an improvement over the existing test suite?

sanderr: Before this change the resource was set up to always fail and this is what the test verified. Adding the `import lorem` did not result in a test failure because the test expected the resource to fail. This change adds a case where the test expects the resource to succeed. I've run this test without the project install fix and it failed, so it is meant to verify that the project install works correctly.

edvgui: Right, I missed that it used to be checking it was failing, not succeeding

sanderr: I missed that for an embarrassingly long time myself, wondering why this test kept succeeding even when I added `raise Exception()` right at the top of the plugin file.

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: I am satisfied indeed! :p

edvgui (approved):

PR #184, opened 2022-02-08
Project install use server env vars

### LINKED ISSUE #183: project install does not take server's env vars into account

sanderr:
# Description

The project install introduced in 1.4.0 did not use the server's env vars configured in `/etc/sysconfig/inmanta-server` into account. This PR runs the installation script inside a transient unit with the same `EnvironmentFile` as the inmanta server.

closes #183 

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -99,8 +99,8 @@ class RemoteOrchestrator:
 
     def export_service_entities(self) -> None:
         """Initialize the remote orchestrator with the service model and check if all preconditions hold"""
-        self.sync_project()
         self._project._exporter.run_export_plugin("service_entities_exporter")
+        self.sync_project()
 
     def _ensure_environment(self) -> None:
         """Make sure the environment exists"""
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -253,7 +253,7 @@ class RemoteOrchestrator:
             )
             shell_script_inline: str = (
                 # use the server's environment variables for the installation
-                "sudo systemd-run User=inmanta EnvironmentFile=/etc/sysconfig/inmanta-server"
+                "sudo systemd-run -p User=inmanta -p EnvironmentFile=/etc/sysconfig/inmanta-server --wait"
                 " /opt/inmanta/bin/python -c %s" % shlex.quote(python_script_inline)
             )
             subprocess.check_output(
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -252,9 +252,8 @@ class RemoteOrchestrator:
                 "project.install_modules();"
             )
             shell_script_inline: str = (
-                "sudo -u inmanta"
-                # set the environment for this command to run in, remove comments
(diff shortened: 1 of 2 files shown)


sanderr:
@wouterdb could you run a quick test when you find the time?

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -245,18 +246,22 @@ def sync_project(self) -> None:
...
+            python_script_inline: str = (
                 "from inmanta.module import Project;"
-                f'project = Project("{server_path}", venv_path="{venv_path}");'
+                f"project = Project('{server_path}', venv_path='{venv_path}');"
                 "project.install_modules();"
             )
+            shell_script_inline: str = (
+                # use the server's environment variables for the installation
+                "sudo systemd-run User=inmanta EnvironmentFile=/etc/sysconfig/inmanta-server"
+                " /opt/inmanta/bin/python -c %s" % shlex.quote(python_script_inline)
```
arnaudsjs, line 257: I don't think this is going to work:

* The user should be provided to `systemd-run` via the `--uid` option and not via `User=inmanta`
* An environment file cannot be passed via `EnvironmentFile=/etc/sysconfig/inmanta-server`. I am not even sure whether these is support to pass an environment file to `systemd-run`
* `systemd-run` returns immediately and not when the command returns. The `--wait` option should be used to make this a blocking call.

arnaudsjs (changes requested):

src/pytest_inmanta_lsm/remote_orchestrator.py
sanderr: I meant to set the options via `-p` (forgot to add it here). I've tested this on 192.168.2.98 (`systemd-run` with `User` and `EnvironmentFile` supplied, not the entire install). Do you believe it should work like that? I was not familiar with `systemd-run` myself so it's possible I misunderstand how to use it, but from the man pages and my experimenting I'd expect this to work.
I'll add the `--wait` option, good catch.

arnaudsjs: Indeed. With `-p` it should work fine.

arnaudsjs (approved):

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -245,18 +246,22 @@ def sync_project(self) -> None:
...
             subprocess.check_output(
                 SSH_CMD
                 + [
                     f"-p {self._ssh_port}",
                     f"{self._ssh_user}@{self.host}",
-                    f"sudo -u inmanta /opt/inmanta/bin/python -c '{install_script_inline}'",
+                    shell_script_inline,
                 ],
                 stderr=subprocess.PIPE,
             )
```
wouterdb, line 267: Can you put this try catch block in:

```python
          except subprocess.CalledProcessError as e:
                LOGGER.error("Process failed out: " + e.output.decode())
                LOGGER.error("Process failed err: " + e.stderr.decode())
                raise
```

wouterdb (changes requested):
the two lines needs to be swapped

wouterdb (approved):

wouterdb:
> @wouterdb you've tested this with the lines swapped?

yes, think so (I'll double check)

wouterdb:
Confirmed!

PR #189, opened 2022-02-15
Issue/186 add support for ssl and auth

### LINKED ISSUE #186: Add support for SSL and authentication
related to https://inmanta.zendesk.com/agent/tickets/82

Introduce environment variables to support authentication and SSL

The code responsible to load this config is here: 
- https://github.com/inmanta/pytest-inmanta-lsm/blob/master/src/pytest_inmanta_lsm/remote_orchestrator.py#L69
Corresponding config code in core:
- https://github.com/inmanta/inmanta-core/blob/master/tests/conftest.py#L712

The expected config variables are

`INMANTA_SSL`, `INMANTA_TOKEN` and `INMANTA_CA_CERT` 

These should set the corresponding options in the sections `compiler_rest_transport` and `client_rest_transport`

Make sure to also update the `Readme.md`

Hugo-Inmanta:
# Description

~~First steps towards adding support for SSL and authentication.
I didn't change anything in the inmanta-core repo even though https://github.com/inmanta/inmanta-core/blob/master/tests/conftest.py#L712 is referenced in the issue~~

Add support for SSL and authentication

closes #186 

# Self Check:

~~- [ ] Changelog entry~~
~~- [ ] Type annotations are present~~
~~- [ ] Code is clear and sufficiently documented~~
~~- [ ] No (preventable) type errors (check using make mypy or make mypy-diff)~~
~~- [ ] Correct, in line with design~~

# Reviewer Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ The following options are available.
  * `--lsm_port` port to use to ssh to the remote orchestrator, overrides INMANTA_LSM_PORT
  * `--lsm_environment` the environment to use on the remote server (it is created if it doesn't exist), overrides INMANTA_LSM_ENVIRONMENT
  * `--lsm_noclean` Don't cleanup the orchestrator after last test (for debugging purposes)
- * `--lsm_ssl` Connect to the remote orchestrator using SSL/TLS
- * `--lsm_token` The token used to authenticate to the remote orchestrator when authentication is enabled.
- * `--lsm_ca_cert` The path to the CA certificate file used to authenticate the remote orchestrator.
+ * `--lsm_ssl` Connect to the remote orchestrator using SSL/TLS, overrides INMANTA_LSM_SSL
+ * `--lsm_token` The token used to authenticate to the remote orchestrator when authentication is enabled, overrides INMANTA_LSM_TOKEN
+ * `--lsm_ca_cert` The path to the CA certificate file used to authenticate the remote orchestrator, overrides INMANTA_LSM_CA_CERT
  
--- a/README.md
+++ b/README.md
@@ -72,3 +72,6 @@ The following options are available.
  * `--lsm_port` port to use to ssh to the remote orchestrator, overrides INMANTA_LSM_PORT
  * `--lsm_environment` the environment to use on the remote server (it is created if it doesn't exist), overrides INMANTA_LSM_ENVIRONMENT
  * `--lsm_noclean` Don't cleanup the orchestrator after last test (for debugging purposes)
+ * `--ssl` Use SSL
+ * `--token` The token to use when using SSL
+ * `--ca_cert` The certificate to use when performing authentication
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ The following options are available.
  * `--lsm_port` port to use to ssh to the remote orchestrator, overrides INMANTA_LSM_PORT
  * `--lsm_environment` the environment to use on the remote server (it is created if it doesn't exist), overrides INMANTA_LSM_ENVIRONMENT
  * `--lsm_noclean` Don't cleanup the orchestrator after last test (for debugging purposes)
- * `--ssl` Use SSL
- * `--token` The token to use when using SSL
- * `--ca_cert` The certificate to use when performing authentication
+ * `--lsm_ssl` Connect to the remote orchestrator using SSL/TLS
+ * `--lsm_token` The token used to authenticate to the remote orchestrator when authentication is enabled.
+ * `--lsm_ca_cert` The path to the CA certificate file used to authenticate the remote orchestrator.
+ 
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -120,7 +120,7 @@ def remote_orchestrator(project: Project, request, remote_orchestrator_settings)
 
     if ssl:
         if not os.path.isfile(ca_cert):
-            raise FileNotFoundError("Invalid path to CA certificate file")
+            raise FileNotFoundError(f"Invalid path to CA certificate file: {ca_cert}")
         ca_cert = os.path.abspath(ca_cert)
     else:
         if ca_cert:
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -122,6 +122,10 @@ def remote_orchestrator(project: Project, request, remote_orchestrator_settings)
         if not os.path.isfile(ca_cert):
             raise FileNotFoundError("Invalid path to CA certificate file")
         ca_cert = os.path.abspath(ca_cert)
+    else:
+        if ca_cert:
+            LOGGER.warning("ssl option is set to False, so the CA certificate won't be used")
+
 
     # set the defaults here and lets the fixture override specific values
     settings: Dict[str, Union[bool, str, int]] = {
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -126,7 +126,6 @@ def remote_orchestrator(project: Project, request, remote_orchestrator_settings)
         if ca_cert:
             LOGGER.warning("ssl option is set to False, so the CA certificate won't be used")
 
-
     # set the defaults here and lets the fixture override specific values
     settings: Dict[str, Union[bool, str, int]] = {
         "auto_deploy": True,
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -115,8 +115,8 @@ def remote_orchestrator(project: Project, request, remote_orchestrator_settings)
     port = get_opt_or_env_or(request.config, "inm_lsm_remote_port", "22")
     noclean = get_opt_or_env_or(request.config, "inm_lsm_noclean", "false").lower() == "true"
     ssl = get_opt_or_env_or(request.config, "inm_lsm_ssl", "false").lower() == "true"
-    token = get_opt_or_env_or(request.config, "inm_lsm_token")
-    ca_cert = get_opt_or_env_or(request.config, "inm_lsm_ca_cert")
+    token = get_opt_or_env_or(request.config, "inm_lsm_token", None)
+    ca_cert = get_opt_or_env_or(request.config, "inm_lsm_ca_cert", None)
 
     # set the defaults here and lets the fixture override specific values
     settings: Dict[str, Union[bool, str, int]] = {
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -36,9 +36,9 @@ option_to_env = {
     "inm_lsm_remote_port": "INMANTA_LSM_PORT",
     "inm_lsm_env": "INMANTA_LSM_ENVIRONMENT",
     "inm_lsm_noclean": "INMANTA_LSM_NOCLEAN",
-    "inm_ssl": "INMANTA_SSL",
-    "inm_token": "INMANTA_TOKEN",
-    "inm_ca_cert": "INMANTA_CA_CERT",
+    "inm_lsm_ssl": "INMANTA_LSM_SSL",
+    "inm_lsm_token": "INMANTA_LSM_TOKEN",
+    "inm_lsm_ca_cert": "INMANTA_LSM_CA_CERT",
 }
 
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -72,3 +72,6 @@ The following options are available.
  * `--lsm_port` port to use to ssh to the remote orchestrator, overrides INMANTA_LSM_PORT
  * `--lsm_environment` the environment to use on the remote server (it is created if it doesn't exist), overrides INMANTA_LSM_ENVIRONMENT
  * `--lsm_noclean` Don't cleanup the orchestrator after last test (for debugging purposes)
+ * `--ssl` Use SSL
+ * `--token` The token to use when using SSL
+ * `--ca_cert` The certificate to use when performing authentication
```
edvgui, line 77: We prefix all the argument related to this pytest plugin with `lsm_`, could you add it for those new arguments as well?  (Same for the environment variables, we use `INMANTA_LSM_...`)

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -66,6 +69,21 @@ def pytest_addoption(parser):
...
     )
+    group.addoption(
+        "--ssl",
+        dest="inm_ssl",
+        help="Use SSL",
+    )
+    group.addoption(
+        "--token",
+        dest="inm_token",
+        help="the token to use when using SSL, overrides INMANTA_TOKEN",
```
edvgui, line 80:
```suggestion
        help="The token to use when using SSL, overrides INMANTA_TOKEN",
```

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -66,6 +69,21 @@ def pytest_addoption(parser):
...
+    )
+    group.addoption(
+        "--ca_cert",
+        dest="inm_ca_cert",
+        help="the certificate to use when performing authentication, overrides INMANTA_CA_CERT",
```
edvgui, line 85:
```suggestion
        help="The certificate to use when performing authentication, overrides INMANTA_CA_CERT",
```

edvgui (changes requested):

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -57,20 +60,31 @@ def __init__(
...
-        inmanta_config.Config.set("client_rest_transport", "port", "8888")
+
+        for section in ["compiler_rest_transport", "client_rest_transport"]:
+            inmanta_config.Config.set(section, "host", host)
+            inmanta_config.Config.set(section, "port", "8888")
+
+            # Config for SSL and authentication:
+            inmanta_config.Config.set(section, "inm_ssl", ssl)
+            inmanta_config.Config.set(section, "inm_token", token)
+            inmanta_config.Config.set(section, "inm_ca_cert", ca_cert)
```
arnaudsjs, line 87: The above-mentioned code sets incorrect config options. You can find the correct values: 
* https://github.com/inmanta/inmanta-core/blob/4312f9d0c124c1cb710c9039c05afce52f5af1b6/tests/conftest.py#L712
Or
* https://docs.inmanta.com/community/dev/reference/config.html#client_rest_transport.ssl

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -45,6 +45,9 @@ def __init__(
         settings: Dict[str, Union[bool, str, int]],
         noclean: bool,
         ssh_port: str = "22",
+        ssl: str = "false",
```
arnaudsjs, line 48: I think it would be good to use the type `bool` here. That way the type checker will enforce that we don't pass anything else but `"True"` or `"False"` in this method and also into the `inmanta_config.Config.set()` method.

arnaudsjs (changes requested):
We should also add a test case to verify that this actually works. If we don't, it will break at some point. @wouterdb Do we have an orchestrator available with SSL and authentication enabled that can be used for a test?

README.md
arnaudsjs: @Hugo-Inmanta The rationale behind this prefixing is the following: `pytest-inmanta-lsm` is a pytest plugin. This means that these config options will be available for use in the `pytest` command. It's important that we don't get a config between the config options we define in this plugin and the config options defined by other pytest plugins. As such we prefix all config option with `lsm_`, that highly reduce the change of a naming conflict.

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -45,6 +45,9 @@ def __init__(
         settings: Dict[str, Union[bool, str, int]],
         noclean: bool,
         ssh_port: str = "22",
+        ssl: bool = False,
+        token: str = "admin",
+        ca_cert: str = "admin",
```
arnaudsjs, line 50: Default values?

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -96,6 +114,9 @@ def remote_orchestrator(project: Project, request, remote_orchestrator_settings)
     user = get_opt_or_env_or(request.config, "inm_lsm_remote_user", "centos")
     port = get_opt_or_env_or(request.config, "inm_lsm_remote_port", "22")
     noclean = get_opt_or_env_or(request.config, "inm_lsm_noclean", "false").lower() == "true"
+    ssl = get_opt_or_env_or(request.config, "inm_lsm_ssl", "false").lower() == "true"
+    token = get_opt_or_env_or(request.config, "inm_lsm_token", None)
+    ca_cert = get_opt_or_env_or(request.config, "inm_lsm_ca_cert", None)
```
arnaudsjs, line 119: The typing of the `get_opt_or_env_or()` method has to be adjusted to allow `None` as a default value.

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -57,20 +60,31 @@ def __init__(
         :param settings: The inmanta environment settings that should be set on the remote orchestrator
         :param noclean: Option to indicate that after the run clean should not run. This exposes the attribute to other
                         fixtures.
+        :param ssl: Option to indicate wether SSL should be used or not. Defaults to false
```
arnaudsjs, line 63:
```suggestion
        :param ssl: Option to indicate whether SSL should be used or not. Defaults to false
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -57,20 +60,31 @@ def __init__(
...
-        inmanta_config.Config.set("compiler_rest_transport", "port", "8888")
-        inmanta_config.Config.set("client_rest_transport", "host", host)
-        inmanta_config.Config.set("client_rest_transport", "port", "8888")
+
+        for section in ["compiler_rest_transport", "client_rest_transport"]:
+            inmanta_config.Config.set(section, "host", host)
+            inmanta_config.Config.set(section, "port", "8888")
+
+            # Config for SSL and authentication:
+            inmanta_config.Config.set(section, "ssl", ssl)
```
arnaudsjs, line 85:
```suggestion
            inmanta_config.Config.set(section, "ssl", str(ssl))
```

README.md
```diff
@@ -72,3 +72,7 @@ The following options are available.
  * `--lsm_port` port to use to ssh to the remote orchestrator, overrides INMANTA_LSM_PORT
  * `--lsm_environment` the environment to use on the remote server (it is created if it doesn't exist), overrides INMANTA_LSM_ENVIRONMENT
  * `--lsm_noclean` Don't cleanup the orchestrator after last test (for debugging purposes)
+ * `--lsm_ssl` Connect to the remote orchestrator using SSL/TLS
```
arnaudsjs, line 75: We also need to add some explanation saying how this option should be used. Intuitively I thought this was a flag, but this option has a value as well. We should indicate which values are valid.

README.md
```diff
@@ -72,3 +72,7 @@ The following options are available.
...
+ * `--lsm_token` The token used to authenticate to the remote orchestrator when authentication is enabled.
+ * `--lsm_ca_cert` The path to the CA certificate file used to authenticate the remote orchestrator.
```
arnaudsjs, line 77: Let's also mention the corresponding environment variables.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -44,28 +47,43 @@ def pytest_addoption(parser):
...
     )
     group.addoption(
         "--lsm_noclean",
         dest="inm_lsm_noclean",
         help="Don't cleanup the orchestrator after tests (for debugging purposes)",
     )
+    group.addoption(
+        "--lsm_ssl",
+        dest="inm_lsm_ssl",
+        help="Connect to the remote orchestrator using SSL/TLS",
```
arnaudsjs, line 75:
```suggestion
        help="Connect to the remote orchestrator using SSL/TLS, overrides INMANTA_LSM_SSL",
```

arnaudsjs (changes requested):

src/pytest_inmanta_lsm/plugin.py
arnaudsjs, line 119: * We should also raise an exception when `ca_cert` references a non-existing file. 
* It would also be good to translate `ca_cert` to an absolute path. That way changing the `cwd` somewhere in the code will not result in undesired behavior.

arnaudsjs (changes requested):

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -96,6 +114,17 @@ def remote_orchestrator(project: Project, request, remote_orchestrator_settings)
     user = get_opt_or_env_or(request.config, "inm_lsm_remote_user", "centos")
     port = get_opt_or_env_or(request.config, "inm_lsm_remote_port", "22")
     noclean = get_opt_or_env_or(request.config, "inm_lsm_noclean", "false").lower() == "true"
+    ssl = get_opt_or_env_or(request.config, "inm_lsm_ssl", "false").lower() == "true"
+    token = get_opt_or_env_or(request.config, "inm_lsm_token", None)
+    ca_cert = get_opt_or_env_or(request.config, "inm_lsm_ca_cert", None)
+
+    if ssl:
+        if not os.path.isfile(ca_cert):
+            raise FileNotFoundError("Invalid path to CA certificate file")
```
arnaudsjs, line 123:
```suggestion
            raise FileNotFoundError(f"Invalid path to CA certificate file: {ca_cert}")
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -44,7 +44,10 @@ def __init__(
         project: Project,
         settings: Dict[str, Union[bool, str, int]],
         noclean: bool,
+        token: Optional[str],
+        ca_cert: Optional[str],
         ssh_port: str = "22",
+        ssl: bool = False,
```
arnaudsjs, line 50: This is a breaking change on the API of the `RemoteOrchestrator` class. I am not sure whether we use this object in other places than the `pytest-inmanta-lsm` library. It's probably best to play it safe and to make sure that this change is backwards compatible.
```suggestion
        ssh_port: str = "22",
        token: Optional[str] = None,
        ca_cert: Optional[str] = None,
        ssl: bool = False,
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -57,20 +60,34 @@ def __init__(
...
+            inmanta_config.Config.set(section, "host", host)
+            inmanta_config.Config.set(section, "port", "8888")
+
+            # Config for SSL and authentication:
+            if ssl:
+                inmanta_config.Config.set(section, "ssl", str(ssl))
+                if token:
+                    inmanta_config.Config.set(section, "token", token)
+                if ca_cert:
+                    inmanta_config.Config.set(section, "ssl_ca_cert_file", ca_cert)
```
arnaudsjs, line 90: SSL and authentication can be enabled/disabled independently on the Inmanta server. SSL is enabled by setting the `server.ssl_cert_file` and the `server.ssl_key_file` config option. Authentication is enabled via the `server.auth` config option. As such, `pytest-inmanta-lsm` should be able to enable/disable SSL and authentication independently as well.

arnaudsjs (changes requested):
We should also mention this change in the `CHANGELOG.md` file.

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -117,7 +134,7 @@ def ensure_project(project_name: str) -> str:
             result = client.project_list()
             assert (
                 result.code == 200
-            ), f"Wrong reponse code while verifying project, got {result.code} (expected 200): \n{result}"
+            ), f"Wrong reponse code while verifying project, got {result.code} (expected 200): \n{result.result}"
```
edvgui, line 137: Not something you did, but while you are there, could you also fix the assertion message below? By adding the `.result` as you did here

src/pytest_inmanta_lsm/remote_orchestrator.py
Hugo-Inmanta: I'm not sure I understand what you mean here.

arnaudsjs: With the above-mention code it's not possible to make `pytest-inmanta-lsm` talk to an Inmanta orchestrator that has authentication enabled, but not SSL, because the `inmanta_config.Config.set(section, "token", token)` line is present under the `if ssl` statement. It would be good to decouple them. Does that make things more clear?

```suggestion
            if token:
                inmanta_config.Config.set(section, "token", token)
            if ssl:
                inmanta_config.Config.set(section, "ssl", str(ssl))
                if ca_cert:
                    inmanta_config.Config.set(section, "ssl_ca_cert_file", ca_cert)
```

Hugo-Inmanta:
Is this alright for the CHANGELOG.md ?
```suggestion
# v 1.5.0 (?)
Changes in this release:

# v 1.4.2 (2022-02-25)
Changes in this release:
- Add support for SSL and anthentication (#186)

# v 1.4.1 (2022-02-10)
Changes in this release:
- Run project installation with server's environment variables
```

arnaudsjs:
> 

This repository still keeps it's changelog in a plain CHANGELOG.md file. In these repos, the top section is always the section for the next minor release. In this specific case, the last release was `v1.4.1` and the next minor release (unreleased version) is `v1.5.0`. This PR implements a new feature. According to the rules of semantic versioning, this change needs to be released with a minor version increment compared to the last release. As such, this change should go into the `v1.5.0` section. It's not required to replace the `?` with a date. The release tooling will take care of that.

edvgui (approved):

arnaudsjs (approved):

edvgui:
@Hugo-Inmanta merge-tool doesn't watch this repo, you will have to press the green button :wink:

PR #194, opened 2022-02-28
Issue/192 support iso5 containers

### LINKED ISSUE #192: Add support for iso5 in a container
Using pytest-inmanta-lsm with an iso5 orchestrator running in a container will raise the following error:
```
FAILED tests/test_service.py::test_create_service[nokia-45] - subprocess.CalledProcessError: Command '['ssh', '-o', 'StrictHostKeyChecking=no', '-o', 'UserKnownHostsFile=/dev/null', '-p 2222', 'inmanta@192.168.2.16', 'sudo systemd-run -p User=inmanta -p EnvironmentFile=...
```

This is because systemd is not running in the container.  To fix this, we should introduce a env variable that allows the user of pytest-inmanta-lsm to skip the loading of env variable using `systemd-run` as any ssh session opened in the container will already have all the required env vars loaded.

The env var documentation should clearly state that when it is set it is considered the responsibility of the user to ensure the host exposese the server's env vars to the ssh connection.

edvgui:
# Description

Add option to tell pytest-inmanta-lsm it will be reaching an orchestrator in a container, which doesn't have a systemd process running in it.

closes #192 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/pytest_inmanta_lsm/managed_service_instance.py
+++ b/src/pytest_inmanta_lsm/managed_service_instance.py
@@ -321,7 +321,7 @@ class ManagedServiceInstance:
                     # Version is not given, so version does not need to be verified
                     return True
                 elif current_state.version not in desired_versions:
-                    raise VersionMismatchError(self, desired_versions, current_state.version)
+                    raise VersionMismatchError(self, desired_versions, current_state.version or 0)
                 else:
                     return True
             elif (
--- a/src/pytest_inmanta_lsm/managed_service_instance.py
+++ b/src/pytest_inmanta_lsm/managed_service_instance.py
@@ -321,7 +321,7 @@ class ManagedServiceInstance:
                     # Version is not given, so version does not need to be verified
                     return True
                 elif current_state.version not in desired_versions:
-                    raise VersionMismatchError(self, desired_versions, current_state.version or 0)
+                    raise VersionMismatchError(self, desired_versions, current_state.version)
                 else:
                     return True
             elif (
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -74,9 +74,9 @@ def pytest_addoption(parser):
         "--lsm_container_env",
         dest="inm_lsm_container_env",
         help=(
-            "If set to true, expect the the orchestrator to be running in a container without systemd.  "
+            "If set to true, expect the orchestrator to be running in a container without systemd.  "
             "It then assumes that all environment variables required to install the modules are loaded into "
-            "each ssh session automatically."
+            "each ssh session automatically.  Overrides INMANTA_LSM_CONTAINER_ENV."
         ),
     )
     group.addoption(
@@ -164,10 +164,10 @@ def remote_orchestrator(project: Project, request, remote_orchestrator_settings)
         project=project,
         settings=settings,
         noclean=noclean,
-        container_env=container_env,
         ssl=ssl,
         token=token,
         ca_cert=ca_cert,
+        container_env=container_env,
     )
     remote_orchestrator.clean()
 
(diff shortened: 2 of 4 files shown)


edvgui:
I tested this on my machine with an iso5 container env and it seems to be working.  I didn't test it against a non-container env without the extra variable set on my machine but the ci did.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -69,15 +70,30 @@ def pytest_addoption(parser):
         dest="inm_lsm_noclean",
         help="Don't cleanup the orchestrator after tests (for debugging purposes)",
     )
+    group.addoption(
+        "--lsm_container_env",
+        dest="inm_lsm_container_env",
+        help=(
+            "If set to true, expect the the orchestrator to be running in a container without systemd.  "
```
arnaudsjs, line 77:
```suggestion
            "If set to true, expect the orchestrator to be running in a container without systemd.  "
```

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -69,15 +70,30 @@ def pytest_addoption(parser):
         dest="inm_lsm_noclean",
         help="Don't cleanup the orchestrator after tests (for debugging purposes)",
     )
+    group.addoption(
```
arnaudsjs, line 73: Also mention the environment variable that can be used to set this setting to the help message.

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         settings: Dict[str, Union[bool, str, int]],
         noclean: bool,
         ssh_port: str = "22",
+        container_env: bool = False,
```
arnaudsjs, line 48: It would be safer to add this new argument at the end of the argument list. Adding it here breaks the API of the `RemoteOrchestrator` class when an instance is created using positional arguments. I don't know whether this is an issue, as I don't know whether this class is used in another library as well, but I don't see any reason why we would break the API if we don't have to.

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -70,6 +71,7 @@ def __init__(
         self._ssh_port = ssh_port
         self._settings = settings
         self.noclean = noclean
+        self.container_env = container_env
```
arnaudsjs, line 74: Also explain this argument in the docstring.

src/pytest_inmanta_lsm/managed_service_instance.py
```diff
@@ -321,7 +321,7 @@ def compare_states(current_state: State, wait_for_states: List[str]) -> bool:
                     # Version is not given, so version does not need to be verified
                     return True
                 elif current_state.version not in desired_versions:
-                    raise VersionMismatchError(self, desired_versions, current_state.version)
+                    raise VersionMismatchError(self, desired_versions, current_state.version or 0)
```
arnaudsjs, line 324: Do you know why `current_state.version` is Optional? What is the semantic meaning of `current_state.version=None`. When I look at the code it seems like `current_state.version is None` can never happen.

arnaudsjs (changes requested):

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: Good point

src/pytest_inmanta_lsm/managed_service_instance.py
edvgui: I think this is because in the version comparison mechanism, None means whatever value the other has is a match of mine.  I simply did it to make mypy happier, but I will remove it

arnaudsjs (approved):

PR #199, opened 2022-03-29
Issue/also report on getting stuck

wouterdb:
# Description

Currently, we don't get sufficient error output when lifecycles get stuck, this PR attempts to improve that. 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 Changes in this release:
 - Add support for iso5 container environment (#192)
 - Add support for SSL and anthentication (#186)
+- Report skipped and deploying resources when reaching a bad lifecycle state
 
 # v 1.4.1 (2022-02-10)
 Changes in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 Changes in this release:
 - Add support for iso5 container environment (#192)
 - Add support for SSL and anthentication (#186)
-- Report skipped and deploying resources when reaching a bad lifecycle state
+- Report skipped and deploying resources when reaching a bad lifecycle state (#199)
 
 # v 1.4.1 (2022-02-10)
 Changes in this release:
--- a/src/pytest_inmanta_lsm/failed_resources_logs.py
+++ b/src/pytest_inmanta_lsm/failed_resources_logs.py
@@ -34,9 +34,8 @@ class FailedResourcesLogs:
         for resource in get_version_result["resources"]:
             resource_id = resource["resource_id"]
 
-            # Only interested in failed resources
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 Changes in this release:
 - Add support for iso5 container environment (#192)
 - Add support for SSL and anthentication (#186)
+- Report skipped and deploying resources when reaching a bad lifecycle state
```
Hugo-Inmanta, line 5: Maybe add issue number for consistency:
```suggestion
- Report skipped and deploying resources when reaching a bad lifecycle state (#199)
```

src/pytest_inmanta_lsm/failed_resources_logs.py
```diff
@@ -34,9 +34,8 @@ def _extract_logs(self, get_version_result: Dict[str, Any]) -> List[Tuple[str, s
         for resource in get_version_result["resources"]:
             resource_id = resource["resource_id"]
 
-            # Only interested in failed resources
-            if resource["status"] != "failed":
-                continue
+            if resource["status"] in ["deploying", "skipped", "failed"]:
```
arnaudsjs, line 37: This class is only instantiated at https://github.com/inmanta/pytest-inmanta-lsm/blob/afe9d375a9ae20105227440aee6375d7478eb82d/src/pytest_inmanta_lsm/managed_service_instance.py#L352. That method explicitly defines a collection of bad_states. Why don't we check `resource["status"]` against those bad states instead of comparing it against the above-mentioned hard coded values?

arnaudsjs (changes requested):

src/pytest_inmanta_lsm/failed_resources_logs.py
wouterdb: The bad_states there are service states like "up" or "delete_validating_failed"

arnaudsjs: I see.

arnaudsjs (approved):

src/pytest_inmanta_lsm/failed_resources_logs.py
Hugo-Inmanta, line 37: I'm not very familiar with this part of the project, but it might be better for maintainability to use an Enum ?

Hugo-Inmanta (approved):

src/pytest_inmanta_lsm/failed_resources_logs.py
wouterdb: in general yes
This enum exist in core

I did it like this because it was like this.

PR #213, opened 2022-05-12
Issue/docker orchestrator

edvgui:
# Description

pytest-inmanta-lsm can now deploy a local orchestrator in a container, if one of our official image is used.

Blocked by https://github.com/inmanta/pytest-inmanta/pull/289

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


README.md
```diff
@@ -63,24 +63,75 @@ def test_full_cycle(project, remote_orchestrator):
...
+```
+pytest-inmanta-lsm:
+  --lsm-ca-cert=LSM_CA_CERT
+                        The path to the CA certificate file used to authenticate the remote
+                        orchestrator. (overrides INMANTA_LSM_CA_CERT)
+  --lsm-container-env   If set to true, expect the orchestrator to be running in a container
+                        without systemd.  It then assumes that all environment variables required
+                        to install the modules are loaded into each ssh session automatically.
+                        (overrides INMANTA_LSM_CONTAINER_ENV)
+  --lsm-doc-orch        If set, the fixtures will deploy and orchestrator on the host, using docker
```
wouterdb, line 80: I don't know if doc is a good short hand, as it usually refers to documentation. 

(but can't think of anything better)

README.md
```diff
@@ -112,3 +163,33 @@ export INMANTA_MODULE_REPO=https://USER:LICENSE_TOKEN@modules.inmanta.com/git/in
     pytest tests
 ```
 
+### Deploy a local orchestrator
+
+It is possible to deploy an orchestrator locally and run the tests against it.  The orchestrator will be deployed as a container, using docker.  Here are the prerequisites in order to make it work:
```
wouterdb, line 168: Could you add the exact/minimal command to use this to improve copy-pastability.

setup.py
```diff
@@ -23,9 +23,21 @@ def read(fname):
...
+            "resources/docker-compose.yml",
+            "resources/my-env-file",
+            "resources/my-server-conf.cfg",
+            "py.typed",
+        ]
+    },
+    include_package_data=True,
+    install_requires=[
+        "pytest-inmanta~=2.3",
+        "docker-compose~=1.29",
```
wouterdb, line 38: I wonder if we should make this optional or always just pull in the whole docker-python toolchain?

src/pytest_inmanta_lsm/docker_orchestrator.py
```diff
@@ -0,0 +1,234 @@
...
+        self.project.up(detached=True)
+
+    def _down(self) -> None:
+        self.project.down(
+            remove_image_type=ImageType.none,
+            include_volumes=True,
+        )
+
+    def __enter__(self) -> "DockerOrchestrator":
+        self._cwd = Path(mkdtemp())
```
wouterdb, line 200: For performance reasons, it might be interesting to be able to not remove everything every time and to not always use a random location.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -122,26 +148,42 @@ def remote_orchestrator_settings() -> Dict[str, Union[str, int, bool]]:
...
-            LOGGER.warning("ssl option is set to False, so the CA certificate won't be used")
+        # If the orchestrator is running in a container we deployed ourself, we overwrite
+        # a few configuration parameters with what matches the deployed orchestrator
+        # If the container image behaves differently than assume, those value won't work,
+        # no mechanism exists currently to work around this.
+        ssh_user = "inmanta"
+        ssh_port = "22"
+        container_env = True
+
+    ssl = inm_lsm_ssl.resolve(request.config)
```
wouterdb, line 177: SSL or AUTH + Docker will never work I think?

Perhaps raise a warning?

src/pytest_inmanta_lsm/docker_orchestrator.py
edvgui: Well, you can use the no clean option the first time, then use specify the ip of the container as the host

src/pytest_inmanta_lsm/plugin.py
edvgui: I added a warning if ssl is set, the orchestrator container is used and the docker-compose file is not overwritten

setup.py
edvgui: Not sure, is there a clean way of doing this?

README.md
edvgui: I added more commands, is it better?

README.md
edvgui: I changed the naming, no mode doc or docker, I used `ctr` (short for container), I also removed `orch/orchestrator` as we already have lsm in there

setup.py
edvgui: We could probably add it in another pytest extension (pytest-inmanta-lsm-ctr) but that would makes things complicated no?

PR #221, opened 2022-05-25
Issue/118 use diagnose endpoint

### LINKED ISSUE #118: Use diagnose endpoint on error
Currently pytest-inmanta-lsm contains logic to lookup the reason for a failure. The server offers a more advanced version that can be used by the fixtures.

When the server offers this api use this, if it does not support it yet fall back to the current version.

edvgui:
# Description

Use diagnose endpoint to generate failed validation/deployment reports.  
Deprecated the methods which were previously used for this.

closes #118 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/pytest_inmanta_lsm/managed_service_instance.py
+++ b/src/pytest_inmanta_lsm/managed_service_instance.py
@@ -11,6 +11,8 @@ from pprint import pformat
 from typing import Any, Collection, Dict, List, Optional, Union
 from uuid import UUID
 
+from inmanta_lsm.diagnose.model import FullDiagnosis
+
 from pytest_inmanta_lsm import remote_orchestrator as r_orchestrator
 from pytest_inmanta_lsm.exceptions import VersionExceededError, VersionMismatchError
 from pytest_inmanta_lsm.wait_for_state import State, WaitForState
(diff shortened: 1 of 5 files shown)


src/pytest_inmanta_lsm/managed_service_instance.py
```diff
@@ -349,22 +350,19 @@ def check_start_state(current_state: State) -> bool:
...
-
-            LOGGER.info("No validation failure message, getting failed resource logs")
-
-            # No validation failure message, so getting failed resource logs
-            failed_resource_logs = FailedResourcesLogs(
-                self.remote_orchestrator.client,
-                self.remote_orchestrator.environment,
+            assert result.code == 200, (
+                f"Wrong response code while trying to get the service diagnostic, got {result.code} (expected 200):\n"
+                f"{json.dumps(result.result or {}, indent=4)}"
```
arnaudsjs, line 362: The ticket mentions that a fallback should be done to the old code when the diangose API endpoint is not available. That is not implemented here, but I guess it's also not requires, because all ISO versions have the diagnose endpoint now?

arnaudsjs (approved):

src/pytest_inmanta_lsm/managed_service_instance.py
edvgui: Iso3 might not have it but we don't run any tests against any iso3 orchestrator anyway

PR #236, opened 2022-07-13
Issue/234 only use sudo over ssh when required

### LINKED ISSUE #234: Only use sudo over ssh when required
It should be possible to disable the use of sudo over ssh when the remote user has the correct permissions.

The goal of this ticket is to only use sudo when the remote user is not inmanta. So is _ssh_user is not `inmanta`, only then sudo needs to be in the command string: https://github.com/inmanta/pytest-inmanta-lsm/blob/master/src/pytest_inmanta_lsm/remote_orchestrator.py#L194

If I am not here anymore, @edvgui can also review this.

Hugo-Inmanta:
# Description

only use sudo over ssh when required

closes #234 
# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -161,6 +161,11 @@ class RemoteOrchestrator:
             result.code == 200
         ), f"Wrong response code while creating environment, got {result.code} (expected 200): \n{result.result}"
 
+    def use_sudo(self) -> str:
+        if self._ssh_user == "inmanta":
+            return ""
+        return "sudo "
+
     def sync_project(self) -> None:
         """Synchronize the project to the lab orchestrator"""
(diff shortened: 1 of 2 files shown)


src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -186,13 +191,16 @@ def sync_project(self) -> None:
         remote_path = f"{self._ssh_user}@{self.host}:{server_path}"
         cache_path = f"{server_path[0:-1]}_cache"  # [0:-1] to get trailing slash out of the way!
 
+        # Disable sudo over ssh when the remote user has the correct permissions
+        use_sudo: str = self.use_sudo()
```
Hugo-Inmanta, line 195: Is it ok to only check once? Or should it be done before each ssh invocation ?

PR #243, opened 2022-08-12
Issue/fix module v2 installation

edvgui:
# Description

Work around the failed editable install for some module v2.  This is not the ideal behavior, I am not sure whether we should persist this fix to make it more robust, or rather find the root cause of the editable install failure.  

This should probably be discussed newt week.

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # v 1.9.0 (?)
 Changes in this release:
 
-- We now automatically install all V2 modules found in de library path in editable mode on the remote orchestrator
+- We now automatically install all V2 modules found in de library path on the remote orchestrator from source.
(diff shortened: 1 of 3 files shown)


sanderr (changes requested):
The core team concluded that it would be best not to merge a patch for pytest-inmanta-lsm specifically for `setuptools<64` support. inmanta/inmanta#4737 should fix this behavior even for that version but it will of course not be available until the next release. Until then, pytest-inmanta-lsm for v2 modules with `use-modules-in-place` requires either a dev version of `inmanta-core` or a `setuptools>=64`, which requires a manual update if using the stable RPMs.

The `.decode` fix should be merged in of course.

edvgui:
Closing this in favor of https://github.com/inmanta/pytest-inmanta-lsm/pull/245.

PR #249, opened 2022-09-06
[FEAT] port venv cleanup fixture fixes from core

Hugo-Inmanta:
# Description

Port venv cleanup fixture fixes from core

part of https://github.com/inmanta/pytest-inmanta/issues/318

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,7 +22,7 @@ def set_cwd(testdir):
     pytest_inmanta.plugin.CURDIR = os.getcwd()
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="function", autouse=True)
 def deactive_venv():
     old_os_path = os.environ.get("PATH", "")
     old_prefix = sys.prefix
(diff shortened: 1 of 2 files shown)


tests/test_partial.py
```diff
@@ -32,7 +32,7 @@ def module_venv(pytestconfig) -> abc.Iterator[env.VirtualEnv]:
 
 @pytest.fixture(scope="function")
 def module_venv_active(
-    deactive_venv: None,
+    deactive_venv: abc.Iterator[None],
```
arnaudsjs, line 35: I don't think this is correct. The return type of the fixture is `abc.Iterator[None]`, but the value of the argument here is `None`, isn't it?

tests/conftest.py
```diff
@@ -22,17 +22,26 @@ def set_cwd(testdir):
     pytest_inmanta.plugin.CURDIR = os.getcwd()
 
 
-@pytest.fixture(scope="function", autouse=True)
+@pytest.fixture(scope="function")
```
arnaudsjs, line 25: I don't think we can remove the `autouse` here, can we?

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #281, opened 2023-01-19
Cleanup fixtures pytest inmanta backports

sanderr:
# Description

I noticed that the cleanup fixtures in `pytest-inmanta-lsm` were outdated relative to those of `pytest-inmanta`. Since it builds on top of `pytest-inmanta` I think (not 100% sure) all its fixtures are relevant here as well so I set that straight.

This will require a release of `pytest-inmanta`, which I'll take care of if/once approved.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 inmanta-lsm
-pytest-inmanta
+pytest-inmanta==2.5.0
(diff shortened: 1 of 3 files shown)


wouterdb (approved):
I'm not certain of the finer details of the reset, but I assume it is good?

sanderr:
> but I assume it is good

I think so

arnaudsjs (approved):

PR #284, opened 2023-01-27
Issue/fix lsm mocked iso4

edvgui:
# Description

Fixed bug for lsm mocked test and iso4

# Self Check:

- [x] Changelog entry

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ def read(fname):
 
 setup(
     name="pytest-inmanta-lsm",
-    version="1.10.1",
+    version="1.11.0",
     python_requires=">=3.6",  # also update classifiers
     author="Inmanta",
     author_email="code@inmanta.com",
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ def read(fname):
 
 setup(
     name="pytest-inmanta-lsm",
-    version="1.11.0",
+    version="1.10.1",
     python_requires=">=3.6",  # also update classifiers
     author="Inmanta",
     author_email="code@inmanta.com",
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -9,6 +9,7 @@ import functools
 import json
 import typing
 import uuid
+import warnings
 
 import inmanta.config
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -31,7 +31,7 @@ def read(fname):
 
 setup(
     name="pytest-inmanta-lsm",
-    version="1.11.0",
```
sanderr, line 34: I don't think you want to commit that change here, this should only be done on the next branch.

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -209,7 +209,13 @@ def compile(
...
             m.setenv(inmanta_lsm.const.ENV_INSTANCE_ID, str(service_id))
             m.setenv(inmanta_lsm.const.ENV_INSTANCE_VERSION, str(service.version))
-            m.setenv(inmanta_lsm.const.ENV_PARTIAL_COMPILE, str(self.partial_compile))
+
+            try:
+                m.setenv(inmanta_lsm.const.ENV_PARTIAL_COMPILE, str(self.partial_compile))
+            except AttributeError:
+                # This attribute only exists for iso5+, iso4 doesn't support partial compile.
+                # We then simply don't set the value.
+                pass
```
sanderr, line 218: Shouldn't we raise an exception if it's set to `True`?

edvgui: I think an exception would be a bit too much, a warning would be enough?

sanderr: Fine by me

setup.py
edvgui: I am struggling a bit with the release process for this package.  What we merge as a patch should not get a `1.11.0.dev...` version but rather a `1.10.1.dev...` (while waiting for the actual release), which is why I made this change, and I was indeed planning on changing it once more in the next branch to bring it back to `1.10.0` so that this bumpversion action would bring it to the correct version: https://github.com/inmanta/irt/blob/master/ci/inmanta-tools/Jenkinsfile-pyp#L201

Is there a better way of doing this?

sanderr (approved):

PR #292, opened 2023-02-27
Add option to specify project and env names

cukiernick:
# Description

Feature requested on slack. New cli options are added, allowing to specify environment and project names. Name options are applied only on creation of missing entries (project/env), no modification is done if uuid of environment matches one of the existing environments.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v 1.12.0 (?)
 Changes in this release:
+- Add option to specify project and environment names
 
 # v 1.11.0 (2023-02-20)
 Changes in this release:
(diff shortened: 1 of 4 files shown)


sanderr (changes requested):
Overall the implementation looks good to me. The only thing I'm missing is a note in the help strings and in the `RemoteOrchestrator` docstring that state that the project and environment names will be used when a new environment is created, but ignored if an environment with the given id already exists.

sanderr (approved):
Could you add a similar message to the help string of the two new options so that it's clear to the end user?

PR #301, opened 2023-03-31
Update LSM project mock to include ServiceEntityBindingV2

jptrindade:
# Description

The `lsm_services_update_attributes_v2` was not being mocked resulting in the `test_lsm_mocked` tests on `athonet_mpn` to fail with the update to `ServiceEntityBindingV2`

# Self Check:

~~- [ ] Attached issue to pull request~~
- [x] Changelog entry
- [ x ] Type annotations are present
- [ x ] Code is clear and sufficiently documented
- [ x ] No (preventable) type errors (check using make mypy or make mypy-diff)
- [ x ] Correct, in line with design

### DIFF
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -103,6 +103,14 @@ class LsmProject:
             self.lsm_services_list,
             raising=False,
         )
+
+        self.monkeypatch.setattr(
+            inmanta_plugins.lsm.global_cache.get_client(),
+            "lsm_services_update_attributes",
+            self.lsm_services_update_attributes,
+            raising=False,
+        )
+
         self.monkeypatch.setattr(
             inmanta_plugins.lsm.global_cache.get_client(),
             "lsm_services_update_attributes_v2",
@@ -191,7 +199,7 @@ class LsmProject:
     ) -> inmanta.protocol.common.Result:
         """
         This is a mock for the lsm api, this method is called during allocation to update
-        the attributes of a service.
+        the attributes of a V2 service.
         """
         # Making some basic checks
         service = self.services[str(service_id)]
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -18,7 +18,7 @@ import inmanta_lsm.const
 import inmanta_lsm.model
 import pytest
 import pytest_inmanta.plugin
-from inmanta_lsm.dict_path import DictPath
+from inmanta_lsm.dict_path import DictPath, to_path
 
 # Error message to display when the lsm module is not reachable
 INMANTA_LSM_MODULE_NOT_LOADED = (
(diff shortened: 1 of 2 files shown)


src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -179,6 +187,43 @@ def lsm_services_update_attributes(
...
+        # * When no candidate attributes were set, the new candidate attributes will be equal to the active
+        #   attributes with the attribute updates applied.
+        # * When candidate attributes were set, the update will be applied to the existing candidate
+        #   attributes.
+        if service.candidate_attributes is None:
+            service.candidate_attributes = copy.deepcopy(service.active_attributes)
+
+        attrs = {}
+        for patch in edit:
+            attrs[patch.target] = patch.value
```
wouterdb, line 220: This works for simple paths, for longer paths, this should use dictpath.

wouterdb (changes requested):

src/pytest_inmanta_lsm/lsm_project.py
edvgui: You might be able to simply steal this logic here: https://github.com/inmanta/inmanta-lsm/blob/39e9319381ce6cfc9fd22549e2b5a9cc7128ded2/src/inmanta_lsm/model.py#L2794 (Calling the method with None instead of self should be fine, as this self parameter is not accessed there)

edvgui: Ah no, it does access self: https://github.com/inmanta/inmanta-lsm/blob/39e9319381ce6cfc9fd22549e2b5a9cc7128ded2/src/inmanta_lsm/model.py#L2836

So copying the logic above that line should be fine

wouterdb (approved):

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -179,6 +188,51 @@ def lsm_services_update_attributes(
...
+        #   attributes with the attribute updates applied.
+        # * When candidate attributes were set, the update will be applied to the existing candidate
+        #   attributes.
+        if service.candidate_attributes is None:
+            service.candidate_attributes = copy.deepcopy(service.active_attributes)
+
+        # Edit logic derived from:
+        # https://github.com/inmanta/inmanta-lsm/blob/39e9319381ce6cfc9fd22549e2b5a9cc7128ded2/src/inmanta_lsm/model.py#L2794
+
+        from inmanta_plugins.lsm.allocation_v2 import dict_path
```
edvgui, line 222: You shouldn't use this, but instead the dict_path library from inmanta_lsm directly

edvgui (changes requested):

src/pytest_inmanta_lsm/lsm_project.py
edvgui: https://github.com/inmanta/inmanta-lsm/blob/master/src/inmanta_lsm/dict_path.py

edvgui: Not this: https://github.com/inmanta/lsm/blob/master/plugins/allocation_v2/dict_path.py
(it is in the end the same code, but the second one is clearly just a backward compatibility fix)

jptrindade:
Is the changelog correct? I remember that usually (at rest) there should be one release without a date and without any entry, but this one already had an entry (with no date). So can I just add the new change in the existing release?

edvgui (approved):
> Is the changelog correct? I remember that usually (at rest) there should be one release without a date and without any entry, but this one already had an entry (with no date). So can I just add the new change in the existing release?

It looks good yes

Issue #89, opened 2020-12-09
Environment creation only works if no other environment has been automatically created before


edvgui:
https://github.com/inmanta/pytest-inmanta-lsm/blob/master/src/pytest_inmanta_lsm/remote_orchestrator.py#L125
```python
result = client.create_environment(
    project_id=ensure_project("pytest-inmanta-lsm"),
    name="pytest-inmanta-lsm",
    environment_id=self._env,
)
```
Instead of always using `pytest-inmanta-lsm` as project and environment name (and fail if it is taken), the method `_ensure_environment` should have a smarter approach, which allows to create environment dynamically more easily.

edvgui:
Fixed in https://github.com/inmanta/pytest-inmanta-lsm/pull/312

PR #312, opened 2023-04-19
Issue/299 rsync editable installs

### LINKED ISSUE #299: rsync all editable installed v2 modules to remote orchestrator
The `remote_orchestrator_project_shared` adds the current module to the libs directory if it is a v2 module. The same logic could be applied to any v2 modules that are installed in editable mode in the current environment. I think it would be desirable to do so in all cases.

edvgui:
# Description

1. Sync **all** modules source available to the local project, **except** v2 modules which are not editable installs.
2. Refactor environment configuration
3. Make the remote orchestrator "standalone" (doesn't rely on any `pytest_inmanta` magic, but simply on the remote orchestrator mechanisms)

Closes #299 and #89 

# Self Check:

- [x] Changelog entry


sanderr (commented):
As discussed

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -366,36 +425,68 @@ def sync_project_folder(self) -> None:
...
+            synced_modules.add(module.name)
+
+        # Make sure all the modules we synced appear to be version controlled
+        mkdir_module = ["mkdir"] + [x for module in synced_modules for x in ["-p", module + "/.git"]]
+        self.run_command(mkdir_module, cwd=str(libs_path))
+
+        # Delete all modules which are on the remote libs folder but we didn't sync
+        grep_extra = ["grep", "-v"] + [x for module in synced_modules for x in ["-e", module]]
+        grep_extra_cmd = shlex.join(grep_extra)
+        clear_extra = f"rm -rf $(ls . | {grep_extra_cmd} | xargs)"
```
sanderr, line 487: Are you aware that you can pass a command to xargs? I think `find` even supports that (`-exec`), which could make this a single command instead of 4, not requiring `shell=True`. Not a strict change requirement.

src/pytest_inmanta_lsm/plugin.py
sanderr, line 1: I think the bulk of the block in `remote_orchestrator_shared` that puts the v2 module in the libs dir can go now. Perhaps we should just have it raise an error if the current module is v2 but not installed in editable mode? Then your changes should take care of the rest. It's a breaking change but this is a major release after all.

edvgui: I didn't remove it completely as it can provide helpful information to the user as to why something is not working.  But its behavior has been changed and modules v2 which are not installed in editable mode won't be synced.

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: I manage to avoid the `$(...)` trick, but I don't see how to keep it only one command?

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -366,36 +425,68 @@ def sync_project_folder(self) -> None:
...
+            )
 
-        self.run_command([clear_extra], shell=True, cwd=libs_path)
+            synced_modules.add(module.name)
+
+        # Make sure all the modules we synced appear to be version controlled
+        mkdir_module = ["mkdir"] + [x for module in synced_modules for x in ["-p", module + "/.git"]]
+        self.run_command(mkdir_module, cwd=str(libs_path))
+
+        # Delete all modules which are on the remote libs folder but we didn't sync
```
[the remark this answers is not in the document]
edvgui: To avoid any side effect of a module we don't need that would be installed on the remote orchestrator, I updated the comment

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -366,36 +425,68 @@ def sync_project_folder(self) -> None:
...
-        libs_path = str(self.remote_project_path / "libs")
+                synced_modules.add(module.name)
 
-        if len(modules) > 0:
-            # Make sure all the modules we synced appear to be version controlled
-            mkdir_module = ["mkdir"] + [x for module in modules for x in ["-p", module + "/.git"]]
-            self.run_command(mkdir_module, cwd=libs_path)
+        # Sync all the modules in editable mode which were not found in the project module's paths
+        for module in modules.values():
+            if module.name in synced_modules:
```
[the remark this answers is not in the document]
edvgui: Indeed, the first loop can go now, I kept it only to be able to sync the current v2 module that might not be installed in editable mode.  But given that it is simply not supported anymore, this loop can go

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -35,80 +36,193 @@
...
+                **updates,
+            )
+
+            assert result.code in range(200, 300), str(result.result)
+            return inmanta.data.model.Environment(**result.result["data"])
+        else:
+            return current_environment
+
+    def __str__(self) -> str:
+        return str(self.id)
```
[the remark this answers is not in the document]
edvgui: You are right, I tried to gain some time because I didn't want to update the usage of the environment attribute everywhere, but I think a better solution is simply to keep the old environment attribute and add a new one that contains our new object.

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -35,80 +36,193 @@
...
+                project_id=self.configure_project(client).id,
+            )
+            assert result.code in range(200, 300), str(result.result)
+            return inmanta.data.model.Environment(**result.result["data"])
+
+        current_project = self.get_project(client)
+
+        updates: dict[str, object] = dict()
+        if self.name is not None and current_environment.name != self.name:
+            # We care about the environment name is it is not a match
```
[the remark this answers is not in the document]
edvgui: Fixed

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -366,36 +425,68 @@ def sync_project_folder(self) -> None:
...
 
-        self.run_command([clear_extra], shell=True, cwd=libs_path)
+            synced_modules.add(module.name)
+
+        # Make sure all the modules we synced appear to be version controlled
+        mkdir_module = ["mkdir"] + [x for module in synced_modules for x in ["-p", module + "/.git"]]
+        self.run_command(mkdir_module, cwd=str(libs_path))
+
+        # Delete all modules which are on the remote libs folder but we didn't sync
+        grep_extra = ["grep", "-v"] + [x for module in synced_modules for x in ["-e", module]]
```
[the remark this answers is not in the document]
edvgui: To be honest I struggle to understand the alternative solution you propose.  What I want to do is a simple flatmap, and to me the most easy way to achieve this in python is via list comprehension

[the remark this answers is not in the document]
sanderr: `itertools.chain.from_iterable` is basically flatten, though I admit it's verbose (alas Python doesn't have a `flatmap` method). I personally find it more readable than the nested list comprehension even so, but that might be because I just replace it with flatten in my head. I find it easier to parse. But like I said, this is a very subjective thing, I'm completely fine to leave it as is if you find the list comprehension approach more readable.

src/pytest_inmanta_lsm/remote_orchestrator.py
sanderr: `find . -maxdepth 1 \( -name one -o -name two \) -exec echo {} +`, replace `echo` with `rm -rf` and `one`, `two` with the module names.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -234,37 +232,50 @@ def remote_orchestrator_shared(
                     "The module being tested is a v1 module but it is also installed as a v2 module. Local compiles will use"
                     "the v2, but only the v1 will by synced to the server."
```
sanderr, line 233: This is not necessarily true anymore: if the v2 is installed in editable mode it will be synced to the server.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -234,37 +232,50 @@ def remote_orchestrator_shared(
...
                 )
+
+        elif installed is None:
+            # Make sure that the module v2 source can be found, it should always be the case, unless
+            # this fixture is used outside of the context of a module test suite.
+            raise RuntimeError(f"Module {mod.name}'s source could not be found")
```
sanderr, line 239: This scenario is a nice catch, looks like an oversight in the initial implementation. But I think a more appropriate error message would be that the module is not installed, this informs the user how to resolve it. Honestly I'd just pack this condition together with the next one and raise the same error, it's accurate enough: not installed => not installed in editable mode.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -219,9 +218,8 @@ def remote_orchestrator_shared(
     Shared project to be used by the remote orchestrator. Ensures the module being tested is synced to the remote orchestrator
     even if it is a v2 module.
     """
-    in_place = inm_mod_in_place.resolve(request.config)
-    # no need to do anything if this version of inmanta does not support v2 modules or if in_place already adds it to the path
-    if hasattr(module, "ModuleV2") and not in_place:
+    # no need to do anything if this version of inmanta does not support v2 modules
+    if hasattr(module, "ModuleV2"):
```
sanderr, line 222: Perhaps it would be useful for readability purposes to extract this block into a `verify_v2_install` method or something? I probably should have done this in the first place back when I implemented this. Not a strict change request.

tests/test_partial.py
```diff
@@ -21,12 +21,13 @@
 from inmanta import env
 
 
-@pytest.fixture(scope="session")
-def module_venv(pytestconfig) -> abc.Iterator[env.VirtualEnv]:
+@pytest.fixture(scope="function")
```
sanderr, line 24: Doesn't this greatly increase test runtime?

tests/utils.py
```diff
@@ -65,6 +65,7 @@ def module_v2_venv(module_path: str) -> Iterator[env.VirtualEnv]:
                 "-X",
                 "module",
                 "install",
+                "-e",
```
sanderr, line 68: This change scares me somewhat. There are subtle differences between the normal install and the editable one that are mostly irrelevant in normal operation but that cause tricky issues when done in-process. I need to give this some more thought. I do see why you do it of course, I guess we'll probably have to, but we do need to consider potential implications. Then again, if the tests succeed it's probably fine. I'll get back to you on this.

sanderr: Let's run with it, I think it's fine.

tests/test_partial.py
sanderr: I do see why you made the change, I'm just wondering if we should think about another workaround. I've got a few in mind, all of them hacks though.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -282,25 +277,31 @@ def remote_orchestrator_shared(
...
+            return option.resolve(request.config)
+        except ParameterNotSetException:
+            return None
+
+    token = get_optional_option(inm_lsm_token)
+    environment_name = get_optional_option(inm_lsm_env_name)
+    project_name = get_optional_option(inm_lsm_project_name)
 
     remote_orchestrator = RemoteOrchestrator(
+        module.Project(project_shared._test_project_dir),
```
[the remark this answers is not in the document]
sanderr: Conclusion: I would definitely use `Project.get()` because `Project` has a bunch of parameters that are set by `pytest-inmanta` and may be changed in the future. The venv being key among them.

As to when to call it, the most stable approach would be to do so dynamically because pytest-inmanta creates a new one for each compile. If you think this would be undesirable I can give the potential consequences some more thought but I think we'd need to step very carefully.

src/pytest_inmanta_lsm/remote_orchestrator.py
sanderr: I forgot to negate it. I think a `\!` before the first bracket should do (`\! \( ..`) but I haven't got the man page handy at the moment.

src/pytest_inmanta_lsm/plugin.py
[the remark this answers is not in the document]
edvgui: I updated the logic this way.  Is there a specific error I could catch and a good explanation I could give when `Project.get()` doesn't "work"?  Given that it is not something that the user provides, it looks like something it does not have control over (at least through the `RemoteOrchestrator` class) so giving some pointers on what is going on and how it could be fixed could be nice.

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: Looks like the negation negates the whole parenthesis, what we want is to negate each statement:
```console
(pytest-inmanta-lsm) guillaume@latitude-5411:/tmp$ find . -maxdepth 1 \! \( -name .Test-unix -name .font-unix \) | grep -e "-unix"
./.X11-unix
./.ICE-unix
./.XIM-unix
./.Test-unix
./.font-unix
(pytest-inmanta-lsm) guillaume@latitude-5411:/tmp$ find . -maxdepth 1 \! -name .Test-unix \! -name .font-unix | grep -e "-unix"
./.X11-unix
./.ICE-unix
./.XIM-unix
```

src/pytest_inmanta_lsm/plugin.py
edvgui: Indeed, fixed

src/pytest_inmanta_lsm/plugin.py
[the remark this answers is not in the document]
sanderr: Alas it doesn't raise an exception if no project has been configured, instead it creates one and sets it. You could check for `project.loaded`. I think all calls into the remote orchestrator expect a project to be loaded already, right?

src/pytest_inmanta_lsm/plugin.py
edvgui: Good idea, done

src/pytest_inmanta_lsm/plugin.py
edvgui: I refactored a bit the install check for modules v2, I am raising a common exception, and still keeping separate branches with different error logs as it has already helped me debugging what was going on in the test cases of this project

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -430,40 +435,9 @@ def sync_project_folder(self) -> None:
...
-                if module is None:
-                    LOGGER.warning("%s is not a valid module, it will be skipped", str(module_path))
-                    continue
-
-                remote_module_path = libs_path / module.name
-                self.sync_local_folder(module_path, remote_module_path, excludes=[])
-
-                synced_modules.add(module.name)
-
         # Sync all the modules in editable mode which were not found in the project module's paths
```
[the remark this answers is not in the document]
edvgui: Indeed, fixed it

tests/test_partial.py
edvgui: It does make things slower indeed.  This whole PR adds 20 seconds to the test suite's time, I would say it is acceptable?

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -219,9 +218,8 @@ def remote_orchestrator_shared(
     Shared project to be used by the remote orchestrator. Ensures the module being tested is synced to the remote orchestrator
     even if it is a v2 module.
```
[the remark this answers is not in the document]
edvgui: Done

src/pytest_inmanta_lsm/plugin.py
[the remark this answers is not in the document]
edvgui: It is expected indeed, I will check for project.loaded and log a warning if it is not then

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: Oh indeed, sorry

src/pytest_inmanta_lsm/plugin.py
sanderr: Ok, in that case I do think we should make the error message more useful to the end user though (you may have already done so, haven't checked your latest changes yet).

tests/test_partial.py
sanderr: I'm not sure. 20 seconds now is ok, but we haven't got that many v2 module tests yet. Any new tests we add will probably be v2, and at some point we'll have to migrate to v2 when we drop support for v1. So if we can avoid it I think we should. I'll see if I can think of a way to get the best of both worlds.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -204,6 +204,54 @@ def remote_orchestrator_partial(request: pytest.FixtureRequest) -> Iterator[bool
...
+
+    if isinstance(mod, module.ModuleV1):
+        # Module v1 installed, it can't be badly installed, we exit here.
+        return
+
+    # mod object is constructed from the source dir: it does not contain all installation metadata
+    installed: Optional[module.ModuleV2] = module.ModuleV2Source(urls=[]).get_installed_module(None, mod.name)
+
+    # Generic message to help fix any of the problems reported below
+    how_to_fix = (
```
sanderr, line 222: Nice

sanderr (commented):
All good, except possibly for the increased runtime which I have yet to look into.

tests/utils.py
sanderr, line 68: Before merging, could you double check that this works for pytest-inmanta? I think it might not.

If it does work, could you make the change there as well? Otherwise we may have to rethink this. I think we should try to keep them in sync.

sanderr (approved):

PR #313, opened 2023-04-20
Issue/212 remove deprecated options

### LINKED ISSUE #212: Remove deprecated options
If https://github.com/inmanta/pytest-inmanta-lsm/pull/210 goes in, it will make a lot of test options, and the functions used around it obsolete.  A bunch of todo's where left in the code to be removed with this ticket.

This requires to make a major version bump.

edvgui:
# Description

As we move towards the new major version of pytest-inmanta-lsm, it is (I believe) a good time to remove the deprecated parts of the code.  This PR removes all legacy options.

closes #212 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changes in this release:
 - Update caching mechanism, don't keep project venv in between test session.
 - Halt environment after the full test suite, resume it before each test run. (the environment can be left running using `--lsm-no-halt` option)
 - Don't sync local project's cfcache to the remote orchestrator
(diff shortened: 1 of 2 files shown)


sanderr (commented):
Why are we doing a major in the first place? I don't immediately see any backwards incompatible changes.

sanderr (approved):
I'm adding a second reviewer from core (also coincidentally the build master in case this sets things on fire on Monday).

Hugo-Inmanta (approved):

PR #325, opened 2023-05-26
Issue/pytest inmanta 374 exclude pycache

sanderr:
Extension of inmanta/pytest-inmanta#374 to pytest-inmanta-lsm. I think the reasoning applies here as well and we should ignore the Python cache. Wdyt?


edvgui (approved):
I think it won't make a huge difference given that everything in the `.gitignore` file is already skipped and `__pycache__` dirs usually are in there.  But it doesn't hurt I guess.

PR #327, opened 2023-06-12
Fix issue where pip output is not logged on failure pip command

### LINKED ISSUE #326: Allow a user to increase the log level on the setup script
The setup script of `pytest-inmanta-lsm` uses the default log level of inmanta-core, which is INFO (See: https://github.com/inmanta/pytest-inmanta-lsm/blob/c2bdc4a6bc637c2d8d75a929156012eca404d878/src/pytest_inmanta_lsm/resources/setup_project.py#L28). This makes it difficult to debug issues with pip, because the output that pip produces is only logged at DEBUG level (See: https://github.com/inmanta/inmanta-core/blob/128a7f40eaa363b155f83ebe76e6199d83511098/src/inmanta/env.py#LL709C1-L709C11). This issue should make sure that this log level is configurable in pytest-inmanta-lsm to ease debugging.

arnaudsjs:
# Description

This PR makes sure that the output of `pip` is logged when the command fails. 

closes #326

# Self Check:

- [ ] Changelog entry
- [ ] ~~End user documentation is included or an issue is created for end-user documentation~~

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -325,12 +325,12 @@ class RemoteOrchestrator:
                     f"{self.ssh_user}@{self.host}",
                     cmd,
                 ],
-                stderr=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
                 universal_newlines=True,
             )
         except subprocess.CalledProcessError as e:
             LOGGER.error("Failed to execute command: %s", cmd)
-            LOGGER.error("Subprocess exited with code %d: %s", e.returncode, str(e.stderr))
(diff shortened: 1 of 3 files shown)


src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -325,12 +325,12 @@ def run_command(
                     f"{self.ssh_user}@{self.host}",
                     cmd,
                 ],
-                stderr=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
```
wouterdb, line 328: doesn't check_output require a pipe?

wouterdb (approved):

PR #363, opened 2023-11-21
Fix import for pydantic v2

wouterdb:
# Description

Fix bad import

# Self Check:

- [ ] Changelog entry -> patch on unreleased 


edvgui (approved):
Oops

PR #364, opened 2023-11-22
Fix warning dictpath

FloLey:
# Description

Fix warning dictpath as it should now be imported from inmanta

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -27,6 +27,12 @@ INMANTA_LSM_MODULE_NOT_LOADED = (
     "    - If you are using v2 modules: make sure the inmanta-module-lsm is installed in your venv."
 )
 
+# Try to import from inmanta.util.dict_path, if not available pass
+try:
+    from inmanta.util import dict_path
+except ImportError:
+    pass
+
 
 class LsmProject:
     def __init__(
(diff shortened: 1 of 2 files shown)


src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -27,6 +27,12 @@
     "    - If you are using v2 modules: make sure the inmanta-module-lsm is installed in your venv."
 )
 
+# Try to import from inmanta.util.dict_path, if not available, fall back to the deprecated inmanta_lsm.dict_path
+try:
+    from inmanta.util import dict_path
+except ImportError:
+    from inmanta_lsm import dict_path
+
```
FloLey, line 35: @sanderr is this what you meant with the try catch? also I don't think it deserves a change entry, does it?

sanderr: Indeed. Also agreed on the change entry part. Did you verify that the latest iso6 has dict_path in inmanta-lsm?

sanderr (changes requested):
Requesting changes for open discussion

src/pytest_inmanta_lsm/lsm_project.py
FloLey: You are talking about iso6-stable right? It does indeed have the dict path in inmanta-lsm

tests/utils.py
```diff
@@ -61,9 +61,7 @@ def module_v2_venv(module_path: str) -> Iterator[env.VirtualEnv]:
             [
                 venv.python_path,
                 "-m",
-                "inmanta.app",
-                "-X",
-                "module",
+                "pip",
                 "install",
                 "-e",
```
FloLey, line 66: rerequesting review for this, just to make sure

sanderr (approved):

Issue #369, opened 2023-12-12
Better error when missing docker compose


FelixGaudin:
When docker-compose is not installed on the computer the error code shows 
`FileNotFoundError: [Errno 2] No such file or directory: 'docker-compose'`

Maybe suggest to the user to install it.

PR #385, opened 2024-02-27
Better logs when docker-compose is not installed

FelixGaudin:
# Description

Improve error logs when `docker-compose` is not installed.

closes https://github.com/inmanta/pytest-inmanta-lsm/issues/369

# Self Check:

- [x] Changelog entry
- [ ] ~~Type annotations are present~~
- [ ] ~~Sufficient test cases (reproduces the bug/tests the requested feature)~~
- [ ] ~~End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )~~

### DIFF
--- a/src/pytest_inmanta_lsm/orchestrator_container.py
+++ b/src/pytest_inmanta_lsm/orchestrator_container.py
@@ -43,8 +43,10 @@ def run_cmd(*, cmd: List[str], cwd: Path) -> Tuple[str, str]:
             universal_newlines=True,
             env=env_vars,
         )
-    except FileNotFoundError:
-        raise FileNotFoundError("The `docker-compose` command is not found. You need it to have a local orchestrator.")
+    except FileNotFoundError as e:
+        if e.filename == "docker-compose":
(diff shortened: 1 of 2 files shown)


src/pytest_inmanta_lsm/orchestrator_container.py
```diff
@@ -32,16 +32,20 @@ def run_cmd(*, cmd: List[str], cwd: Path) -> Tuple[str, str]:
...
+            cwd=str(cwd),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            encoding="utf-8",
+            text=True,
+            universal_newlines=True,
+            env=env_vars,
+        )
+    except FileNotFoundError:
+        raise FileNotFoundError("The `docker-compose` command is not found. You need it to have a local orchestrator.")
```
edvgui, line 47: I would add an additional check that the missing file is indeed `docker-compose`, because it could be also something else (like the `docker-compose.yml` file)

edvgui (approved):

PR #386, opened 2024-03-01
Issue/async waiters

edvgui:
# Description

- Add `AsyncManagedServiceInstance` class, with asynchronous methods to manage a service instance.
- Add testing helper, to group multiple async testing scenarios into one.

# Self Check:

- [x] Changelog entry


src/pytest_inmanta_lsm/service_instance.py
```diff
@@ -0,0 +1,607 @@
...
+        return sync_call
+
+    def __setattr__(self, __name: str, __value: object) -> None:
+        """
+        Set an attribute on the wrapped service instance.
+        """
+        if __name != "async_service_instance":
+            return setattr(self.async_service_instance, __name, __value)
+        else:
+            super().__setattr__(__name, __value)
```
edvgui, line 607: This one might not be required, I mainly added it for "symmetry", any opinion on it?

src/pytest_inmanta_lsm/service_instance.py
```diff
@@ -0,0 +1,607 @@
...
+            tid=self.remote_orchestrator.environment,
+            service_entity=self.service_entity_name,
+            attributes=attributes,
+            service_instance_id=self._instance_id,
+        )
+        assert (
+            service_instance.version == 1
+        ), f"Error while creating instance: wrong version, got {service_instance.version} (expected 1)"
+
+        # Safe the instance id for later
```
wouterdb, line 354:
```suggestion
        # Save the instance id for later
```

src/pytest_inmanta_lsm/util.py
```diff
@@ -0,0 +1,83 @@
...
+    :contact: code@inmanta.com
+    :license: Inmanta EULA
+"""
+
+import asyncio
+import collections.abc
+import typing
+
+
+class MultiException(RuntimeError):
```
wouterdb, line 14: Using https://docs.python.org/3/library/exceptions.html#ExceptionGroup would be cool, but that is only from 3.11 I don't know which versions we target here?

src/pytest_inmanta_lsm/util.py
```diff
@@ -0,0 +1,83 @@
...
+        self.others = others
+
+    def format(self) -> str:
+        return "Reported %d errors" % len(self.others)
+
+    def __str__(self) -> str:
+        return "Reported %d errors:\n\t" % len(self.others) + "\n\t".join([str(e) for e in self.others])
+
+
+def execute_scenarios(
```
wouterdb, line 28: I would expect an async version of this call to exists as well?

src/pytest_inmanta_lsm/service_instance.py
```diff
@@ -0,0 +1,607 @@
+"""
```
wouterdb, line 1: I might be interesting to make the name of this more distinct wrt to the old one. We want people to clearly understand there are multiple versions, which are (slightly) different

README.md
```diff
@@ -17,21 +17,71 @@ It requires an LSM enabled orchestrator, with no ssl or authentication enabled,
...
+        wait_for_state="up",
+        timeout=60,
+    )
+
+    # Delete the instance
+    await instance.delete(wait_for_state="terminated", timeout=60)
 
-def test_full_cycle(project, remote_orchestrator):
+
+def test_full_cycle(project: plugin.Project, remote_orchestrator: remote_orchestrator.RemoteOrchestrator) -> None:
```
wouterdb, line 79: I would not explain this more sophisticated packing mechanism as the base case. 

I would start with a simple testcase first (either async or not) and then go into the more subtle packing case .

wouterdb:
Looks very nice

src/pytest_inmanta_lsm/service_instance.py
edvgui: Do you have a suggestion for a better name?

README.md
edvgui: I reorganized it a bit, is it better now?

src/pytest_inmanta_lsm/util.py
edvgui: Done

src/pytest_inmanta_lsm/service_instance.py
wouterdb: perhaps put it in an async (or aio) package ? 

I don't really know

edvgui: After a (long) discussion with @arnaudsjs I renamed everything to the following:
1. sync version: `remote_service_instance.RemoteServiceInstance`
2. async version: `remote_service_instance_async.RemoteServiceInstance`

I also added a `.pyi` file for the sync implementation, but there is nothing that ensures it will stay "synced" in the future (updates in `async` file will also need to update the `.pyi` file)

README.md
```diff
@@ -19,19 +19,63 @@ It requires an LSM enabled orchestrator, with no ssl or authentication enabled,
...
+        start_version=created.version,
+        timeout=60,
+    )
+
+    # Delete the instance
+    instance.delete(wait_for_state="terminated", timeout=60)
+```
+> source: [test_quickstart.py::test_transient_state](./examples/quickstart/tests/test_quickstart.py)
+
+For a more advanced test case, you might also want to deploy multiple services.  You could either test them one by one, or, parallelize them, to:
```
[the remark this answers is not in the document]
edvgui: What do you mean with separately? This is already presented as an advance usage scenario, should it be a different use case completely?

wouterdb (approved):

src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -0,0 +1,544 @@
...
+    def __init__(
+        self,
+        instance: T,
+        bad_states: typing.Collection[str],
+        log: model.ServiceInstanceLog,
+        *args: object,
+    ) -> None:
+        super().__init__(
+            instance,
+            f"Service instance for into bad state {log.state} (version: {log.version}) " f"from bad state list: {bad_states}",
```
hlloreda, line 75:
```suggestion
            f"Service instance for into bad state {log.state} (version: {log.version}) f"from bad state list: {bad_states}",
```

src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -0,0 +1,544 @@
...
+                    last_state = log.state
+
+                # Save the current version
+                last_version = log.version
+
+            if time.time() - start > timeout:
+                # We reached the timeout, we should stop waiting and raise an exception
+                LOGGER.info(
+                    "Service instance %s exceeded timeout while waiting for %s, current state is %s",
+                    self._instance_name,
```
hlloreda, line 302: This could be none if we call this method before creating something
But I guess you wanted to do something like this?
```suggestion
                    self.instance_name,
```

src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -0,0 +1,544 @@
...
+                    target_state,
+                    log.state,
```
hlloreda, line 304: Log could be None if the list is empty (once https://github.com/inmanta/inmanta-lsm/issues/1635 is fixed) and the timeout is too short. It should never happen I know, but if we can do something about it

src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -0,0 +1,544 @@
...
+                start_version=current_version,
+            )
+        else:
+            return await self.get()
+
+    async def set_state(
+        self,
+        state: str,
+        *,
+        current_version: typing.Optional[int] = None,
```
hlloreda, line 498: `state` and `current_version` are not part of the doc string

src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -0,0 +1,544 @@
...
+        )
+        if service_entity.service_identity is not None:
+            self._instance_name = (
+                f"{self.service_entity_name}"
+                f"({service_entity.service_identity}={service_instance.service_identity_attribute_value})"
+            )
+            LOGGER.info("Created instance has name %s", self.instance_name)
+        else:
+            # There is no service_identity_display_name, so we use the instance id
+            self._instance_name = f"{self.service_entity_name}({self.instance_id})"
```
hlloreda, line 373: This is already done in the property, do we need this?

hlloreda (changes requested):
Looks great, only minor problems

src/pytest_inmanta_lsm/remote_service_instance_async.py
edvgui: I managed to rely on `last_state`, which will always be set.  And if it is still None, I handle the case with better logging

src/pytest_inmanta_lsm/remote_service_instance_async.py
edvgui: Indeed, I removed it

src/pytest_inmanta_lsm/remote_service_instance_async.py
edvgui: Indeed, I added them

src/pytest_inmanta_lsm/remote_service_instance_async.py
edvgui: Fixed

src/pytest_inmanta_lsm/remote_service_instance_async.py
edvgui: Fixed

src/pytest_inmanta_lsm/util.py
```diff
@@ -0,0 +1,99 @@
...
+
+else:
+    # The ExceptionGroup has been introduced in python3.11, if we have
+    # a recent enough version of python, we should use it.
+    # https://docs.python.org/3/library/exceptions.html#ExceptionGroup
+    ExceptionGroupError = ExceptionGroup
+
+
+async def execute_scenarios(
+    *scenarios: collections.abc.Awaitable,
```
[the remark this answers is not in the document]
edvgui: Fixed

hlloreda (approved):

PR #389, opened 2024-03-12
Halt environment before clearing it

arnaudsjs:
# Description

Make sure to first halt the environment before clearing it.

# Self Check:

- [ ] ~~Attached issue to pull request~~
- [x] Changelog entry
- [ ] ~~End user documentation is included or an issue is created for end-user documentation~~

### DIFF
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -326,7 +326,11 @@ def remote_orchestrator_shared(
     )
 
     # Make sure we start our test suite with a clean environment
+    result = remote_orchestrator.client.halt_environment(remote_orchestrator.environment)
+    assert result.code in range(200, 300), str(result.result)
     remote_orchestrator.clear_environment()
+    result = remote_orchestrator.client.resume_environment(remote_orchestrator.environment)
+    assert result.code in range(200, 300), str(result.result)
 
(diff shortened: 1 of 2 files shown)


src/pytest_inmanta_lsm/plugin.py
```diff
@@ -340,17 +344,17 @@ def get_optional_option(option: StringTestParameter) -> Optional[str]:
...
     remote_orchestrator.setup_config()
 
-    # If --lsm-no-clean is used, leave the orchestrator as it is, with all its
-    # file, otherwise cleanup the project
-    if not remote_orchestrator_no_clean:
-        remote_orchestrator.clear_environment()
-
     # If --lsm-no-halt is used, leave the orchestrator running at the end of the
     # test suite.  Otherwise the environment is halted.
     if not remote_orchestrator_no_halt:
```
edvgui, line 349: Should we then always halt the environment when we don't specify `--lsm-no-clean`? (meaning `--lsm-no-halt` can't be used/has not effect without `--lsm-no-clean`)

arnaudsjs: This PR just become a bit more complicates, because it was decided to implement https://github.com/inmanta/inmanta-core/issues/6587. I am going to unassign you from review for now and rework this PR based on the above-mentioned ticket.

Issue #354, opened 2023-10-06
Add service accepts any service


bartv:
The [add service method](https://github.com/inmanta/pytest-inmanta-lsm/blob/master/src/pytest_inmanta_lsm/lsm_project.py#L238) accepts any service instance.

It should raise exceptions if it does not exist so that you can rely on lsm mocked tests for refactoring services.

edvgui:
Closed by https://github.com/inmanta/pytest-inmanta-lsm/pull/391

PR #391, opened 2024-03-20
Issue/352 add exporter mock

### LINKED ISSUE #352: Add LsmProject method to test lsm export_service_entities plugin
When using the `remote_orchestrator` fixture, tests usually start by running this helper method: https://github.com/inmanta/pytest-inmanta-lsm/blob/8658f3408fd02dd494e6d1202a77965771664321/src/pytest_inmanta_lsm/remote_orchestrator.py#L627

It will run the lsm export plugin to send the service inventory to the orchestrator.  This plugin does some validation on the model that is only done during this export.  We could add a similar helper method to the `LsmProject` class, allowing to test this export plugin without using a real orchestrator.

edvgui:
# Description

- Add `export_service_entities` helper method to mocked lsm project
- Add capability to prefill the default values when performing a validation compile with the mocked lsm project
- Allow to easily reuse model used in `export_service_entities` for all later compiles.
- Validate that any service added to the `LsmProject` object using `add_service` method is part of one of the exported services.

Closes #352 and #354

# Self Check:

- [x] Changelog entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,9 +202,6 @@ def test_model(lsm_project: pytest_inmanta_lsm.lsm_project.LsmProject) -> None:
     # Do a first validation compile, and add all default values to our candidate attributes
     lsm_project.compile(service_id=service.id, validation=True, add_defaults=True)
 
-    # Assert that the default value has been added to our attributes
-    assert "value_with_default" in service.candidate_attributes
-
     # The first validation compile went fine, move to the next state
     pytest_inmanta_lsm.lsm_project.promote(service)
     service.version += 1
--- a/README.md
+++ b/README.md
@@ -193,8 +193,20 @@ def test_model(lsm_project: lsm_project.LsmProject) -> None:
     # for EACH compile
     lsm_project.add_service(service)
 
-    # Run a compile, with central focus the service we just created
+    # Export the service entities
+    lsm_project.export_service_entities("import quickstart")
+
+    # Do a first validation compile, and add all default values to our candidate attributes
+    lsm_project.compile("import quickstart", service_id=service.id, validation=True, add_defaults=True)
+
+    # The first validation compile went fine, move to the next state
+    pytest_inmanta_lsm.lsm_project.promote(service)
+    service.version += 1
+    service.state = "creating"
+
+    # Do a second compile, in the non-validating creating state
     lsm_project.compile("import quickstart", service_id=service.id)
+
 ```
 
 ## Options and environment variables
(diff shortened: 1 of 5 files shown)


examples/quickstart/tests/test_quickstart.py
```diff
@@ -235,8 +240,16 @@ def test_model(lsm_project: lsm_project.LsmProject) -> None:
...
+    # Do a first validation compile, and add all default values to our candidate attributes
+    lsm_project.compile("import quickstart", service_id=service.id, validation=True, add_defaults=True)
+
+    # The first validation compile went fine, move to the next state
+    pytest_inmanta_lsm.lsm_project.promote(service)
+    service.version += 1
+    service.state = "creating"
 
-    # Do a second compile, everything should go fine
+    # Do a second compile, in the non-validating creating state
```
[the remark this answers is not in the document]
edvgui: Good idea, I added the assertion

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -281,6 +282,9 @@ def lsm_service_catalog_get_entity(
         This is a mock for the lsm api, this method is called during export of the
         service entities.
         """
+        assert self.service_entities is not None, "Bad usage!!"
```
jptrindade, line 285: The assert message could give some suggestions on the correct usage

edvgui: Indeed, fixed

jptrindade (approved):

README.md
```diff
@@ -189,12 +196,20 @@ def test_model(lsm_project: lsm_project.LsmProject) -> None:
         service_identity_attribute_value=None,
     )
 
-    # Add the service to the mocked server.  From now on it will be taken into account
-    # for EACH compile
+    # Add a service to our inventory
     lsm_project.add_service(service)
```
bartv, line 200: I am wondering if validation and add default shouldn't be options (or even defaults) in add_service?

bartv: I understand you want to offer the flexibility but in "real" lsm this is also enforced by the api server and not the validation compile

edvgui: I wouldn't make it a default, as it would break our test case where we try to reproduce a specific failure for a given "existing" service (which is not created from scratch.  I can add an option to allow it at least

edvgui: Done in https://github.com/inmanta/pytest-inmanta-lsm/pull/391/commits/7a0d37b1c58d695056bbfe0be5d303ec04d4912a

bartv (approved):
cool

PR #392, opened 2024-03-22
Issue/380 partial compile testing helpers

### LINKED ISSUE #380: Help partial compile testing with lsm_project fixture
Testing that a model behaves correctly with partial compile enabled is tricky, but it is possible, and a lot of invariants can be checked, with a fairly standard logic.

The goal of this ticket is to extract such standard logic from this test case: https://code.inmanta.com/solutions/bics/bics_ce/-/blob/47786ed5a4ab88da63fdc4c17e0652b303d1be62/tests/test_lsm_mocked/test_partial_compile.py#L189

`SHARED_RESOURCES` and `OWNED_RESOURCES` should be provided as input as they will vary for each service.

edvgui:
# Description

- Add `LsmProject` helpers to facilitate partial compile testing (#380)
- Add `LsmProject` helpers to facilitate service creation and update:
    - Fill in default attribute values.
    - Determine initial state automatically.
    - Follow the first "auto" state transfers, running the corresponding compiles, and applying the corresponding attribute operations.
 - Change behavior for default values handling introduced in previous MR.  This will now only be done for `create_service` and `update_service`.

closes #380 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -617,7 +617,7 @@ class LsmProject:
 
         if service_entity_name not in self.service_entities:
             raise LookupError(
-                f"Unknown service entity {service_entity_name}.  " f"Known services are: {list(self.service_entities.keys())}."
+                f"Unknown service entity {service_entity_name}.  Known services are: {list(self.service_entities.keys())}."
             )
 
         return self.service_entities[service_entity_name]
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -4,16 +4,22 @@
     :license: Inmanta EULA
 """
 
+import collections
 import copy
 import datetime
 import functools
+import hashlib
 import json
+import logging
+import pathlib
+import re
 import typing
 import uuid
 import warnings
 
 import inmanta.config
 import inmanta.protocol.common
+import inmanta.resources
 import inmanta.util
 import inmanta_lsm.const  # type: ignore[import-not-found]
 import inmanta_lsm.model  # type: ignore[import-not-found]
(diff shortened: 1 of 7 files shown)


src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -414,17 +808,11 @@ def add_service(
...
-            # the default values.
-            self.compile(model=None, service_id=service.id, validation=True, add_defaults=True)
-
     def compile(
         self,
         model: typing.Optional[str] = None,
         service_id: typing.Optional[uuid.UUID] = None,
         validation: bool = True,
-        add_defaults: bool = False,
     ) -> None:
```
edvgui, line 816: These two are noticeable changes which break the (unreleased) behavior introduced yesterday.  I did it for the following reasons:
1. I believe that adding defaults should not be done during (or right before) a compile, as this is not how the api handles it, but rather when the service is created or updated.  We didn't have support for create/update operations yesterday, but we have now.  So I moved this logic in the create/update methods.
2. Given that we now have explicit create/update operations, I think it makes no sense to have the possibility to trigger a validation compile when adding the service, this is actually more the use case of the create method.  So I reverted back that change.

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -462,48 +845,100 @@ def compile(
...
+            service.id,
+            service.service_entity,
+            env,
+        )
+        with self.monkeypatch.context() as m:
+            for k, v in env.items():
+                m.setenv(k, v)
+            self.project.compile(model, no_dedent=False)
+
+    def post_partial_compile_validation(
```
jptrindade, line 878: It would be nice to have a test of this feature too.

edvgui: Good point, I added it

PR #394, opened 2024-03-25
Issue/report state on timeout

edvgui:
# Description

Slightly improve error reporting for remote service instance:
1. Include current state in timeout error
2. Log diagnosis before raising timeout error

# Self Check:

- [ ] ~~Attached issue to pull request~~
- [ ] ~~Changelog entry~~ --> Change on unreleased code

### DIFF
--- a/src/pytest_inmanta_lsm/remote_service_instance_async.py
+++ b/src/pytest_inmanta_lsm/remote_service_instance_async.py
@@ -13,8 +13,8 @@ import typing
 import uuid
 
 import devtools
-from inmanta_lsm import model
-from inmanta_lsm.diagnose.model import FullDiagnosis
+from inmanta_lsm import model  # type: ignore
+from inmanta_lsm.diagnose.model import FullDiagnosis  # type: ignore
 
 from pytest_inmanta_lsm import remote_orchestrator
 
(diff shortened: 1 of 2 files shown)


src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -186,6 +195,24 @@ async def history(self, *, since_version: int = 0) -> list[model.ServiceInstance
             filter={"version": f"ge:{since_version}"},
         )
 
+    async def diagnose(self, *, version: int) -> FullDiagnosis:
```
jptrindade, line 198: Is there a specific reason why we need the `version` to be a kwarg?

jptrindade (approved):

src/pytest_inmanta_lsm/remote_service_instance_async.py
edvgui: It is more a design choice.  I think it is clearer to force this argument name to be provided, as a simple number has very little meaning in a "diagnose" function.  While if you know it is the version, it makes a lot more sense.  It is also aligned with other helper methods on this class (where I made the same design choice).  Do you have an objection about it maybe?  I am not saying that I made the best design choice so I am open for discussion of course

PR #399, opened 2024-03-29
Issue/cleanup agent map

edvgui:
# Description

Since https://github.com/inmanta/inmanta-core/pull/7449, the agent will be started when they are added to the agent map.  This issue has gone unnoticed for so long because this setting of the environment was not reset in between tests.  This PR fixes it.

This will probably require a few labs to be rebuilt.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -429,12 +429,20 @@ def remote_orchestrator(
         "autostart_agent_deploy_interval": 600,
         "autostart_agent_repair_splay_time": 600,
         "autostart_agent_repair_interval": 0,
-        "autostart_agent_map": {
-            "internal": "local:",
-        },
         "lsm_partial_compile": remote_orchestrator_partial,
     }
     settings.update(remote_orchestrator_settings)
+
+    # Get server version and check if the server has the patch which safely allows to reset
+    # the agent map
+    # https://github.com/inmanta/inmanta-core/issues/7448
+    orchestrator_version = version.Version(remote_orchestrator_shared.status.version)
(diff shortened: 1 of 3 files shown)


sanderr (commented):
While I'm all for it in principle, I'm not convinced it's a good idea to add it now. If I understand correctly this will cause tests to start failing for all versions except the latest iso ones. Is that right? If so, I think we should wait for at least a year before we make this change.

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -433,6 +433,17 @@ def remote_orchestrator(
     }
     settings.update(remote_orchestrator_settings)
 
+    # Get server version and check if the server has the patch which safely allows to reset
+    # the agent map
+    # https://github.com/inmanta/inmanta-core/issues/7448
+    orchestrator_version = version.Version(remote_orchestrator_shared.status.version)
+    good_iso7 = orchestrator_version > version.Version("7.1.0.dev")
+    good_iso6 = version.Version("7.dev") > orchestrator_version > version.Version("6.5.0.dev")
```
sanderr, line 441:
```suggestion
    good_iso6 = version.Version("7") > orchestrator_version > version.Version("6.5.0.dev")
```
`7 < 7.dev`

sanderr (changes requested):

src/pytest_inmanta_lsm/plugin.py
edvgui:
```
>>> version.Version("7") < version.Version("7.dev")
False
```

sanderr: You're absolutely right, not sure what I was thinking.

PR #400, opened 2024-03-29
Relax partial compile validation to allow owned-service validation

edvgui:
# Description

The post partial validation logic assumed there would be only one resource set in each partial compile.  This was wrong because:
1. A service can pull other services with it in the partial compile selection logic
2. A service can emit more than one resource set

This PR relaxes a bit the constraint that was set, and checks that the resource set of the service is in present, then gather all the owned resources (from all resource sets).

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -9,6 +9,7 @@ import copy
 import datetime
 import functools
 import hashlib
+import itertools
 import json
 import logging
 import pathlib
@@ -938,12 +939,14 @@ class LsmProject:
         resource_sets = get_resource_sets(self.project)
         if not service.deleted:
             # Check that the only resource set emitted is the one of this service
-            assert resource_sets.keys() == {str(service_id)}
-            _, owned_resources = resource_sets.popitem()
+            assert str(service_id) in resource_sets.keys()
         else:
             # Check that no resource set is emitted
-            assert resource_sets.keys() == set()
-            owned_resources = []
+            assert str(service_id) not in resource_sets.keys()
+
+        # Get all the owned resources, also the ones from other services we
+        # pulled down in our partial selection
+        owned_resources = list(itertools.chain(*resource_sets.values()))
 
         # Check that each resource that is emitted belongs to the expected resource set
         for resource_id in self.project.resources.keys():


src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -938,12 +939,14 @@ def post_partial_compile_validation(
         resource_sets = get_resource_sets(self.project)
         if not service.deleted:
             # Check that the only resource set emitted is the one of this service
-            assert resource_sets.keys() == {str(service_id)}
-            _, owned_resources = resource_sets.popitem()
+            assert str(service_id) in resource_sets.keys()
```
wouterdb, line 942: This doesn't always hold (for systemtenant, we always only emit the resource set of the systemtenant).

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -976,10 +979,21 @@ def post_partial_compile_validation(
         model = pathlib.Path(self.project._test_project_dir, "main.cf").read_text()
         self.project.compile(model)
 
-        # Check that we have as many resource sets as there are services
-        assert get_resource_sets(self.project).keys() == {
-            str(srv.id) for srv in self.services.values() if not srv.deleted and not srv.state == "ordered"
+        # Check each service in the inventory has its own resource set
+        all_resource_sets = get_resource_sets(self.project).keys()
```
wouterdb, line 983: This doesn't always hold (for systemtenant, we always only emit the resource set of the systemtenant).

wouterdb (commented):
I would keep around the validation for the base case (one service-one resource set) and keep a less strict version for more complex scenarios. 

For the complex scenarios, model developers can write their own assertions.

src/pytest_inmanta_lsm/lsm_project.py
edvgui: How about the lifecycle transfers of the different services, are they also in the st resource set?

edvgui:
I changed the focus of the pull request, keeping the current logic, which different modules can replace with their own when needed.  I still added a small helper which gives all the services which are in an exporting state.

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -977,9 +989,7 @@ def post_partial_compile_validation(
         self.project.compile(model)
 
         # Check that we have as many resource sets as there are services
-        assert get_resource_sets(self.project).keys() == {
-            str(srv.id) for srv in self.services.values() if not srv.deleted and not srv.state == "ordered"
-        }
+        assert get_resource_sets(self.project).keys() == self.exporting_services.keys()
```
wouterdb, line 992: I would make it clear in the doc string this only covers the base case

wouterdb (changes requested):

wouterdb (approved):

PR #404, opened 2024-04-10
Issue/facilitate demo setup

edvgui:
# Description

For PoC, we very often don't have a project to clone in the orchestrator environment, we only work with a module, and pytest-inmanta-lsm test cases.  It can be frustrating to clear the environment everytime we want to update the module's code, or service definition.

This MR adds a new fixture, which returns the remote orchestrator object and all its nice features, without the initial cleanup the `remote_orchestrator` fixture does.

# Self Check:

- [ ] Changelog entry


edvgui:
`--lsm-no-halt --lsm-no-clean` will still reset the environment **before** running the test, this doesn't touch the environment at all.

My concrete use case: update a poc environment in which services are already deployed, without having to redeploy these services.  It basically allows to simulate the "export service catalog" button of the web console, without having a git project and a released module.

bartv (approved):

PR #406, opened 2024-04-15
Add the ability to generate a base load similar to the Inmanta dashboard

hlloreda:
# Description

Add a load generator that will contact specific endpoints, one by one, with a configurable time between them

closes solutions/tickets#111

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/load_generator.py
+++ b/src/pytest_inmanta_lsm/load_generator.py
@@ -6,17 +6,13 @@
     :license: Inmanta EULA
 """
 
+import asyncio
 import datetime
-import functools
 import logging
 import threading
 import time
-from dataclasses import dataclass
 
-import inmanta_plugins.restbase.utils
-import requests
-from inmanta_plugins.restbase import handler_logger
-from requests_toolbelt.utils import dump
+from pytest_inmanta_lsm import remote_orchestrator
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


edvgui (changes requested):
I have a few concerns about the current design:
1. pytest-inmanta-lsm is an open-source library, with focus on testing the lsm api.  The restbase module is not something that any user of this library will have access to, it can't be a dependency.
2. The code base of this package/library is centered around the remote orchestrator class, I would keep it this way.  And we also use the inmanta client object, not the request library, let's stay consistent.
3. If I understood correctly, using this will allow you to trigger 30 seconds of load on the orchestrator, then it stops.  I would expect something that runs for the duration of the test (or the context manager)
4. I don't see why you need a custom Thread class, you are the one starting and joining the thread, if any exception occurs, you can handle it already.

Overall, I think it can be made a lot simpler, an infinite loop in a function that goes over all the api calls, with of course a flag that allows to jump out of the loop when required.  This function would need as single argument the remote orchestrator object (and could be wrapped by a contextmanager, that is a good idea).  It is then trivial to make it a fixture if needed, or just use it for localized pieces of tests.

edvgui (changes requested):
This is starting to look more aligned with this code-base, good job!  I think this designs is good, it is going to work, I still have a few questions to challenge your implementations:
1. Why do you need a lock? A lock is useful when multiple processes/thread need to read and write a value, but here only the main thread will write, and the load thread will read.
2. How about restarting the thread? If the object is a context manager, I would expect that you can enter it multiple times, does it work here?
3. What is this thing about refcycle? If that is something to address, my guess would be that there is something wrong in the usage? But maybe I don't understand the issue
4. Do we really want to make all the apis call one after the other, creating bursts on the orchestrator? This is not really what a user would do, is it?
5. What if we exit the context manager, we might have to wait until the end of the loop?

Suggestion for 4 and 5: a custom `sleep` method, which checks if the thread should terminate, and sleeps otherwise.

Another suggestion, which is definitely not required but I would find the idea interesting is to change the definition of your thread class to specifically handle async targets (as well as catching an reporting your error).

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        start_datetime = datetime.datetime.utcnow()
+        nb_datapoints = 15
+        # start_interval and end_interval should be at least <nb_datapoints> minutes separated from each other
+        # But we need also to respect the following constraint:
+        # When round_timestamps is set to True, the number of hours between start_interval and end_interval should be
+        # at least the amount of hours equal to nb_datapoints
+        end_datetime = start_datetime + datetime.timedelta(hours=nb_datapoints + 1)
+
+        while True:
```
wouterdb, line 118: I would spread the load more evenly over the interval: now you are machine-gunning the api at the start of the interval.

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,199 @@
...
+        self.logger.debug("Thread %s has been stopped, closing session", self._thread.name)
+
+    def fetch_thread_status(self) -> bool:
+        """
+        Retrieve the status of whether the thread should keep running or not
+        """
+        with self._lock:
+            return self.SHOULD_THREAD_RUN
+
+    def change_status(self):
```
wouterdb, line 91:
```suggestion
    def stop(self):
```

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,199 @@
...
+        service_type: str,
+        logger: logging.Logger = LOGGER,
+        sleep_time: float = 1.0,
+    ):
+        self.remote_orchestrator = remote_orchestrator
+        self.service_type = service_type
+        self.sleep_time = sleep_time
+        self.logger = logger
+        self._lock = threading.Lock()
+        self.SHOULD_THREAD_RUN = True
```
wouterdb, line 71:
```suggestion
        self.running = True
```

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,199 @@
...
+        self.logger.debug("Starting Thread %s", self._thread.name)
+        self._thread.start()
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        self.change_status()
+        self.logger.debug("Stopping Thread %s", self._thread.name)
+        self._thread.join(30)
+        self.logger.debug("Thread %s has been stopped, closing session", self._thread.name)
+
+    def fetch_thread_status(self) -> bool:
```
wouterdb, line 84:
```suggestion
    def is_running(self) -> bool:
```

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,199 @@
...
+import threading
+import time
+
+from pytest_inmanta_lsm import remote_orchestrator
+
+LOGGER = logging.getLogger(__name__)
+
+
+# Custom Thread Class
+class LoadThread(threading.Thread):
```
wouterdb, line 21: 1. I would make the exception behavior to log it and keep going.  Perhaps keep a boolean if the run was correct for the testcase to assert on
2. I would just put a try except block around the actual code, instead of this fairly complicated custom thread

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,203 @@
...
+    async def remote_call(self, call: typing.Callable[[], typing.Awaitable[model.BaseModel]]) -> None:
+        try:
+            await call()
+        except Exception as e:
+            self.logger.warning("%s - encountered the following error:%s!", self.thread.name, str(e))
+        finally:
+            if not self.is_running():
+                raise LoadException()
+
+            time.sleep(self.sleep_time)
```
edvgui, line 99: For this method, you should rather use `asyncio.sleep`, as the code base is async (even though it is true that blocking the thread is not a big issue, it is cleaner).
Also, it would be good to check `is_running` before calling the api (or after the sleep), and maybe only sleep for small intervals (i.e. half a second) at a time, always checking for `is_running` at each iteration.

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,203 @@
...
+        if self._thread is None:
+            raise RuntimeError("The thread is not initialized, this instance should be used in a context!")
+
+        return self._thread
+
+    def is_running(self) -> bool:
+        """
+        Retrieve the status of whether the thread should keep running or not
+        """
+        return self.running
```
edvgui, line 70: Is this method really useful?

edvgui (changes requested):
We are getting closer to the final implementation, can you add docstrings, type annotations and a test? (also, restbase is still in the dependencies and should go)

edvgui (changes requested):
Can you add docstrings on class/methods and add type annotations on the context manager exit?

edvgui (approved):

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,218 @@
...
+            raise RuntimeError("The thread is not initialized, this instance should be used in a context!")
+
+        return self._thread
+
+    def between_callback(self) -> None:
+        """
+        Will create the load in an async manner. This method will also save any exception that could have occurred in the
+        thread. This mechanism allows to raise an exception once the thread has joined, see the `__exit__` method.
+        """
+        loop = asyncio.new_event_loop()
```
wouterdb, line 87: Why combine asyncio with thread? i.e. is there a specific idea behind this?

wouterdb (changes requested):

src/pytest_inmanta_lsm/load_generator.py
hlloreda: We rely on a thread to not block the main thread from doing anything other than running the test. And we rely on async because we use the `request` method of the remote orchestrator, which is async

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,218 @@
...
+
+    async def create_load(self) -> None:
+        """
+        Create load until `running` flag is set to False
+        """
+        start_datetime = datetime.datetime.utcnow()
+        nb_datapoints = 15
+        # start_interval and end_interval should be at least <nb_datapoints> minutes separated from each other
+        # But we need also to respect the following constraint:
+        # When round_timestamps is set to True, the number of hours between start_interval and end_interval should be
```
wouterdb, line 124: what is round_timestamps?

src/pytest_inmanta_lsm/load_generator.py
```diff
@@ -0,0 +1,218 @@
...
+        finally:
+            await asyncio.sleep(self.sleep_time)
+
+    async def create_load(self) -> None:
+        """
+        Create load until `running` flag is set to False
+        """
+        start_datetime = datetime.datetime.utcnow()
+        nb_datapoints = 15
+        # start_interval and end_interval should be at least <nb_datapoints> minutes separated from each other
```
wouterdb, line 122: this is super confusing. It is about a specific input, not about the actual load generation. This should be made clear

wouterdb (changes requested):
When this is done, it would be nice to have it in the npm scaling tests

wouterdb (approved):

PR #411, opened 2024-04-25
Issue/dump orchestrator on failure

### LINKED ISSUE #409: Add fixture to dump support archive on test failure
It has been implemented for a project already: https://code.inmanta.com/solutions/athonet/athonet_mpn/-/blob/5e342074250f362db7a13c0d2cda5f6fc46b80d9/tests/conftest.py#L914

But ideally we should also:
1. Stream the output, not load it fully in memory: https://github.com/inmanta/inmanta-support/issues/700
2. Allow to turn this behavior on/off with a parameter (default to off), and maybe merge it with the remote fixture logic

edvgui:
# Description

Add `--lsm-dump-on-failure` option, allowing to generate a support archive of the orchestrator when a test fails, and save it in the host /tmp directory.

closes #409 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -14,7 +14,7 @@ import tempfile
 import textwrap
 import time
 import uuid
-from typing import Dict, Generator, Iterator, Optional, Tuple, Union
+from typing import Any, Dict, Generator, Iterator, Optional, Tuple, Union
 from uuid import UUID
 
 import pkg_resources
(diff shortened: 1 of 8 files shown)


src/pytest_inmanta_lsm/plugin.py
```diff
@@ -443,7 +473,42 @@ def remote_orchestrator(
     result = remote_orchestrator_shared.client.resume_environment(remote_orchestrator_shared.environment)
     assert result.code in range(200, 300), str(result.result)
 
-    return remote_orchestrator_shared
+    yield remote_orchestrator_shared
+
+    # Check that status of the test that just ran
+    # https://docs.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures
```
wouterdb, line 479: so fancy!

wouterdb (approved):
I assume you tested this?

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -179,21 +196,6 @@ def remote_orchestrator_host(
...
-            response.raise_for_status()
-        except Exception as exc:
-            LOGGER.warning(str(exc))
-            time.sleep(1)
-            continue
-
-        if response.status_code == 200:
-            return host, port
-
-    raise RuntimeError(f"Couldn't reach the orchestrator at {host}:{port}")
```
edvgui, line 195: I moved this out of this fixture and into the `remote_orchestrator_access` one to be able to use the `RemoteOrchestrator.session` object.  This changes the behavior of the fixture, but not for the expected usage of the tool.

I also changed what is being done in the `RemoteOrchestrator` (environment configuration is now done in the fixture), which is another behavioral change, which I also think is okay as it doesn't change the behavior for the expected usage of the tool.

PR #413, opened 2024-04-29
Issue/containers logs to server.log

edvgui:
# Description

Make sure that containers started by pytest-inmanta-lsm have a server.log file

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/plugin.py
+++ b/src/pytest_inmanta_lsm/plugin.py
@@ -8,6 +8,8 @@
 
 import logging
 import os
+import pathlib
+import re
 import shlex
 import shutil
 import tempfile
@@ -20,6 +22,7 @@ from uuid import UUID
 import pkg_resources
 import pytest
 import pytest_inmanta.plugin
+import pytest_inmanta.test_parameter
 from inmanta import module
 from packaging import version
 from pytest_inmanta.plugin import Project
@@ -132,10 +135,40 @@ def remote_orchestrator_container(
         yield None
         return
 
+    orchestrator_image = inm_lsm_ctr_image.resolve(request.config)
+    latest_compose_file = pathlib.Path(__file__).parent / "resources/docker-compose.yml"
+    legacy_compose_file = pathlib.Path(__file__).parent / "resources/docker-compose-iso6.yml"
+    try:
+        compose_file = inm_lsm_ctr_compose.resolve(request.config)
+    except pytest_inmanta.test_parameter.ParameterNotSetException:
+        # The compose file is not set, we can then either use the default one, or
+        # the default legacy one (for <iso7.1).  To decide which one is the most
+        # appropriate, we parse the container image tag and extract the iso version
+        iso_major_version_match = re.fullmatch(
+            r"(?:.*)\/service-orchestrator:((\d+(?:\.\d+)*)(?:\-dev)?|dev)",
+            orchestrator_image,
+        )
+        if not iso_major_version_match:
+            LOGGER.warning(
+                "Can not parse orchestrator image tag: %s.  Using legacy docker-compose file",
+                orchestrator_image,
+            )
+            compose_file = legacy_compose_file
+        elif iso_major_version_match.group(1) == "dev":
+            # Latest dev, use the the latest compose file
+            compose_file = latest_compose_file
+        elif version.Version(iso_major_version_match.group(2)) >= version.Version("7.1"):
+            # The server has the --db-wait-time option
+            # https://github.com/inmanta/inmanta-core/pull/7217
+            compose_file = latest_compose_file
+        else:
+            # The image is not recent enough, we have to use the legacy docker-compose file
+            compose_file = legacy_compose_file
+
     LOGGER.debug("Deploying an orchestrator using docker")
     with OrchestratorContainer(
-        compose_file=inm_lsm_ctr_compose.resolve(request.config),
-        orchestrator_image=inm_lsm_ctr_image.resolve(request.config),
+        compose_file=compose_file,
+        orchestrator_image=orchestrator_image,
         postgres_version=inm_lsm_ctr_db_version.resolve(request.config),
         public_key_file=inm_lsm_ctr_pub_key.resolve(request.config),
         license_file=inm_lsm_ctr_license.resolve(request.config),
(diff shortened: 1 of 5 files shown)


src/pytest_inmanta_lsm/plugin.py
```diff
@@ -132,10 +135,40 @@ def remote_orchestrator_container(
...
+    latest_compose_file = pathlib.Path(__file__).parent / "resources/docker-compose.yml"
+    legacy_compose_file = pathlib.Path(__file__).parent / "resources/docker-compose-iso6.yml"
+    try:
+        compose_file = inm_lsm_ctr_compose.resolve(request.config)
+    except pytest_inmanta.test_parameter.ParameterNotSetException:
+        # The compose file is not set, we can then either use the default one, or
+        # the default legacy one (for <iso7.1).  To decide which one is the most
+        # appropriate, we parse the container image tag and extract the iso version
+        iso_major_version_match = re.fullmatch(
+            r"(?:.*)\/service-orchestrator:((\d+(?:\.\d+)*)(?:\-dev)?|dev)",
```
sanderr, line 148: Not a strict change request:

```suggestion
            r".*\/service-orchestrator:((\d+(?:\.\d+)*)(?:\-dev)?|dev)",
```
I had to look up what this meant, and I believe it's redundant so might as well keep it  simple?
Personally, I'd even go so far as to drop all of them (replacing with plain `()`) because imo it reads a lot easier.

```suggestion
            r"(?:.*)\/service-orchestrator:((\d+(\.\d+)*)(\-dev)?|dev)",
```

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -132,10 +135,40 @@ def remote_orchestrator_container(
...
+            orchestrator_image,
+        )
+        if not iso_major_version_match:
+            LOGGER.warning(
+                "Can not parse orchestrator image tag: %s.  Using legacy docker-compose file",
+                orchestrator_image,
```
sanderr, line 154: It's not clear to my why we'd default to legacy? All our containers should match this format, shouldn't they?

src/pytest_inmanta_lsm/plugin.py
```diff
@@ -132,10 +135,40 @@ def remote_orchestrator_container(
...
+            )
+            compose_file = legacy_compose_file
+        elif iso_major_version_match.group(1) == "dev":
+            # Latest dev, use the the latest compose file
```
sanderr, line 158:
```suggestion
            # Latest dev, use the the latest compose file, this is always safe because we don't distribute this externally
```

This is not too relevant now, but if we ever want to do the same for iso 7 vs 8, this becomes important, because not everyone has access to the same set of images, so concepts as "latest" are ambiguous. But as long as we only use such a concept for "latest *dev*", we're good because that concept only exists internally, hence everyone (who might use it) does have the same set of images available.

src/pytest_inmanta_lsm/resources/docker-compose.yml
```diff
@@ -9,10 +9,18 @@ services:
...
+      INMANTA_DATABASE_USERNAME: inmanta
+      INMANTA_DATABASE_PASSWORD: inmanta
+      INMANTA_SERVER_BIND_ADDRESS: "0.0.0.0"
+      INMANTA_SERVER_BIND_PORT: 8888
     volumes:
-      - ${INMANTA_LSM_CONTAINER_PUBLIC_KEY_FILE}:/var/lib/inmanta/.ssh/authorized_keys
       - ${INMANTA_LSM_CONTAINER_LICENSE_FILE}:/etc/inmanta/license/com.inmanta.license
       - ${INMANTA_LSM_CONTAINER_ENTITLEMENT_FILE}:/etc/inmanta/license/com.inmanta.jwe
-      - ./my-server-conf.cfg:/etc/inmanta/inmanta.cfg
+      # The last mount is not required, but it is kept for backward compatibility
```
sanderr, line 24: "The last mount" is not very clear imo.

src/pytest_inmanta_lsm/resources/docker-compose-iso6.yml
```diff
@@ -0,0 +1,17 @@
...
+      POSTGRES_PASSWORD: inmanta
+
+  inmanta-server:
+    image: ${INMANTA_LSM_CONTAINER_ORCHESTRATOR_IMAGE}
+    volumes:
+      - ${INMANTA_LSM_CONTAINER_LICENSE_FILE}:/etc/inmanta/license/com.inmanta.license
+      - ${INMANTA_LSM_CONTAINER_ENTITLEMENT_FILE}:/etc/inmanta/license/com.inmanta.jwe
+      - ./my-server-conf.cfg:/etc/inmanta/inmanta.cfg
+      - ./my-env-file:/etc/inmanta/env
+    command: "server --wait-for-host ${INMANTA_LSM_CONTAINER_DB_HOSTNAME} --wait-for-port 5432"
```
sanderr, line 17: It's not clear to me why this command changed, and why the `authorized_keys` file disappeared (I guess the second is a consequence of the former). And I don't have sufficient experience / context to review this part.

sanderr (changes requested):

src/pytest_inmanta_lsm/plugin.py
edvgui: Indeed, I cleaned this up.  I went for named groups, as I find it a lot easier to read when so many groups are in the regex

src/pytest_inmanta_lsm/plugin.py
edvgui: Ours, yes, but nothing prevents the user of pytest-inmanta-lsm to run the tests against a custom image, this is what this case is for.  I added a comment

src/pytest_inmanta_lsm/plugin.py
sanderr: I feel differently, perhaps because I'm so used to generic regexes, rather than Python's dialect. But it's a lot more clear than the non-capturing groups imo (and my objection is very subjective), so fine by me.

src/pytest_inmanta_lsm/resources/docker-compose-iso6.yml
edvgui: I changed it because it is not required anymore, as of pytest-inmanta-lsm==3.3.0, for container orchestrators we deploy ourself, we don't ssh in, we "docker exec" in

sanderr (approved):

PR #414, opened 2024-04-29
Issue/fix wait for transient

edvgui:
# Description

Fix race condition in `RemoteServiceInstance.wait_for_state` that would make it return a `ServiceInstance` with a different version that the one we asked the method to wait for.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changes in this release:
 - Add support to `LsmProject.compile` to have multiple instances selected
 - Add `LoadGenerator` helper to generate some load on the remote orchestrator
 - Add `--lsm-dump-on-failure` option, allowing to generate a support archive of the orchestrator when a test fails, and save it in the host /tmp directory. (#409)
+- Fix race condition in `RemoteServiceInstance.wait_for_state` that would make it return a `ServiceInstance` with a different version that the one we asked the method to wait for.
 
 # v 3.3.0 (2024-04-15)
 Changes in this release:
--- a/src/pytest_inmanta_lsm/remote_service_instance_async.py
+++ b/src/pytest_inmanta_lsm/remote_service_instance_async.py
@@ -247,7 +247,8 @@ class RemoteServiceInstance:
         start_version: int,
     ) -> model.ServiceInstance:
         """
-        Wait for this service instance to reach the desired target state.
+        Wait for this service instance to reach the desired target state.  Returns a ServiceInstance
+        object that is in the state that was waited for.
 
         :param target_state: The state we want to wait our service instance to reach.
         :param target_version: The version the service is expected to be in once we reached the target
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Changes in this release:
 - Add support to `LsmProject.compile` to have multiple instances selected
 - Add `LoadGenerator` helper to generate some load on the remote orchestrator
 - Add `--lsm-dump-on-failure` option, allowing to generate a support archive of the orchestrator when a test fails, and save it in the host /tmp directory. (#409)
+- Fix race condition in `RemoteServiceInstance.wait_for_state` that would make it return a `ServiceInstance` with a different version that the one we asked the method to wait for.
```
sanderr, line 6: Perhaps be a bit more explicit (and less alarming)?
```suggestion
- Fix race condition in `RemoteServiceInstance.wait_for_state` that would make it return a `ServiceInstance` for the latest version rather than the one we asked the method to wait for.
```

src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -24,6 +24,31 @@
...
+    :param log: The ServiceInstanceLog to convert to a ServiceInstance object.
+    :param current: The ServiceInstance to use as a basis for the crafted instance.
+    """
+    try:
+        # Dump the log as a dict, to make it easier to get its attributes values
+        # programmatically
+        raw_log = log.model_dump()
+        return model.ServiceInstance(
+            **{attr: raw_log[attr] if attr in raw_log else value for attr, value in current.model_dump().items()}
+        )
```
sanderr, line 43: This looks OK to me, but if we are planning to add support for something like this natively to inmanta-lsm (for which @wouterdb is probably the better judge), then I don't think forward compatibility is a concern here (we could just call the method if it exists), so I think it would be both easier and cleaner to construct a `ServiceInstance` directly with the appropriate fields.

sanderr (changes requested):
I think an update to the wording of the README that you linked on Slack could be useful as well. It also states something in the sense of (paraphrasing) "we can ensure consistency by passing a version we know we've seen before", where "seen" isn't really what matters, it's more about a version that we know matches the preconditions.

None of my comments are strict change requests. Mostly requesting changes to make sure they aren't overlooked.

src/pytest_inmanta_lsm/remote_service_instance_async.py
edvgui: I took the hint, it is true that it makes it a lot easier to read and understand, I also referenced the corresponding ticket I created in inmanta-lsm: https://github.com/inmanta/pytest-inmanta-lsm/pull/414/commits/350a631f513174d23d7df39014618215e94d5be2

sanderr (approved):

PR #422, opened 2024-06-18
Fix compatibility with latest inmanta-lsm

hlloreda:
# Description

Fix compatibility with updated ServiceInstance model -> new fields:
  - `desired_state_version`
  - `transfer_context`

Refer to inmanta-lsm for more details.

# Self Check:

- [ ] Changelog entry


edvgui:
This won't work for versions of inmanta-lsm <= 7.2 (so all releases so far)

src/pytest_inmanta_lsm/remote_service_instance_async.py
```diff
@@ -52,6 +53,7 @@ def get_service_instance_from_log(log: model.ServiceInstanceLog) -> model.Servic
             deployment_progress=None,
             service_identity_attribute_value=log.service_identity_attribute_value,
             referenced_by=None,
+            transfer_context=log.transfer_context,
```
edvgui, line 56: This is not required, for later versions of inmanta-lsm, which have these new attributes, the `log.to_service_instance` method is already define and should contain the compatibility logic

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # v 3.5.0 (?)
 Changes in this release:
 - Make sure that the orchestrators started by pytest-inmanta-lsm log their output to `/var/log/inmanta/server.log` instead of stdout for rc containers.
+- Fix compatibility with updated ServiceInstance model -> new fields:
+  - `desired_state_version`
+  - `transfer_context`
+
+  Refer to inmanta-lsm for more details.
```
edvgui, line 8: People using the library (which is open source and used by clients) can't refer to inmanta-lsm as it is closed source

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # v 3.7.0 (?)
 Changes in this release:
 - Add support for ng containers.
+- Fix compatibility with updated ServiceInstance model -> new field:
+  - `desired_state_version` : The version of the desired state of a service instance. This version is incremented every time the desired state of the service is changing, i.e. the exported model.
```
edvgui, line 5: I don't understand this sentence.  Also, there is no logic here to actually increment this number?

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -691,16 +691,29 @@ def auto_transfer(self, service_id: uuid.UUID) -> inmanta_lsm.model.ServiceInsta
...
                 return
 
+            try:
+                if is_error_transition:
+                    is_preserving_same_desired_state = transfer.error_same_desired_state
+                else:
+                    is_preserving_same_desired_state = transfer.target_same_desired_state
+
+                if not is_preserving_same_desired_state:
+                    service.desired_state_version += 1
```
edvgui, line 712: Yes, but:
1. I would rather do it at the same place we bump the service version
2. We also need to skip the compile if we don't emit a new desired state version

edvgui (changes requested):

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -671,6 +672,53 @@ def get_service_entity(self, service_entity_name: str) -> inmanta_lsm.model.Serv
...
+        # In case of an `AttributeError`, we don't need to do anything: we are dealing with an old orchestrator
+        is_preserving_same_desired_state = False
+        with suppress(AttributeError):
+            if is_error_transition:
+                is_preserving_same_desired_state = transfer.error_same_desired_state
+            else:
+                is_preserving_same_desired_state = transfer.target_same_desired_state
+
+            if not is_preserving_same_desired_state:
+                service.desired_state_version += 1
```
edvgui, line 709:
```suggestion
        if not is_preserving_same_desired_state:
            service.desired_state_version += 1
```
We also have to do it for the default, right?

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -671,6 +672,53 @@ def get_service_entity(self, service_entity_name: str) -> inmanta_lsm.model.Serv
...
+        Apply this state to our service, if it is different from the current
+        state, also increment the version and the desired state version (if the exported model is changing)
+
+        :param service: The current service
+        :param transfer: The current transfer
+        :param is_error_transition: Is it an error transition?
+        """
+        if is_error_transition:
+            if transfer is None:
+                raise RuntimeError(f"Error transition is not defined for {str(transfer)}!")
```
edvgui, line 691: According to the signature of the method, it can't be None

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -816,10 +854,11 @@ def update_service(
 
         # Go into the update state
         try:
-            service.state = service_entity.lifecycle.get_transfer(
+            transfer = service_entity.lifecycle.get_transfer(
                 from_state=service.state,
                 transfer_type=inmanta_lsm.const.TransferTrigger.ON_UPDATE,
-            ).target
```
hlloreda, line 822: The current implementation is not entirely correct, but I have the feeling that the update path was not correct before. The `version` field was not updated, and we were not doing a recompile, am I missing something?

src/pytest_inmanta_lsm/lsm_project.py
edvgui: I realize now that 2. is not that easy, because in this case, the compile that was called in the method is the one on the current state, not the one triggered after the state transfer.

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -16,6 +16,7 @@
 import typing
 import uuid
 import warnings
+from contextlib import suppress
```
edvgui, line 19:
```suggestion
import contextlib.suppress
```

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -691,39 +739,20 @@ def auto_transfer(self, service_id: uuid.UUID) -> inmanta_lsm.model.ServiceInsta
...
                 service.service_entity,
             )
-            self.compile(
-                service_id=service.id,
-                validation=transfer.validate_,
-            )
             perform_attribute_operation(service, transfer.target_operation)
-            next_state(transfer.target)
+            self.transfer_to_next_state(service=service, transfer=transfer, is_error_transition=False)
         except Exception:
```
edvgui, line 752: This logic is broken, the exception we catch here is from the state afther the transition, nto the one of the current state

src/pytest_inmanta_lsm/lsm_project.py
edvgui: We were doing all of that in the auto_transfer method, but you changed the logic to compile after the transfer, instead of compiling before the transfer, this is very different and indeed breaks everything

edvgui (changes requested):

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -691,39 +739,20 @@ def auto_transfer(self, service_id: uuid.UUID) -> inmanta_lsm.model.ServiceInsta
...
             LOGGER.info(
                 "Triggering compile on state %s before auto transfer (%s) for service %s (%s)",
                 service.state,
                 transfer.description,
                 service.id,
                 service.service_entity,
             )
-            self.compile(
-                service_id=service.id,
-                validation=transfer.validate_,
```
hlloreda, line 719: But the logic here was wrong (at least for the update path), we were compiling for the current state, but with the next transfer, it should have been the transfer we just took right?

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -691,18 +691,30 @@ def auto_transfer(self, service_id: uuid.UUID) -> inmanta_lsm.model.ServiceInsta
...
             if service.state == state:
                 return
 
             service.last_updated = datetime.datetime.now()
             service.version += 1
+
+            try:
+                is_preserving_same_desired_state = (
+                    is_error_transition and transfer.error_same_desired_state
+                ) or transfer.target_same_desired_state
```
edvgui, line 711:
```suggestion
                is_preserving_same_desired_state = transfer.error_same_desired_state if is_error_transition else transfer.target_same_desired_state
```
Isn't this more clear?

edvgui (approved):

PR #449, opened 2024-09-27
Add support for new docker compose command

jptrindade:
# Description

Jenkins runner 11 uses `docker compose` instead of `docker-compose`

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/orchestrator_container.py
+++ b/src/pytest_inmanta_lsm/orchestrator_container.py
@@ -226,7 +226,7 @@ class OrchestratorContainer:
             return ["docker-compose"]
         except FileNotFoundError:
             try:
-                subprocess.run(args=["docker-compose", "--help"])
+                subprocess.run(args=["docker", "compose", "--help"])
                 return ["docker", "compose"]
             except FileNotFoundError:
                 raise FileNotFoundError(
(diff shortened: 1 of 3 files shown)


src/pytest_inmanta_lsm/orchestrator_container.py
```diff
@@ -221,22 +219,37 @@ def orchestrator_ips(self) -> List[IPv4Address]:
...
         return int(self.config.get("server", "bind-port", vars={"fallback": "8888"}))
 
+    @property
+    def docker_compose(self) -> list[str]:
+        try:
+            subprocess.run(args=["docker-compose", "--help"])
+            return ["docker-compose"]
+        except FileNotFoundError:
+            try:
+                subprocess.run(args=["docker-compose", "--help"])
```
edvgui, line 229:
```suggestion
                subprocess.run(args=["docker", "compose", "--help"])
```

edvgui (approved):
One copy-paste error, looks good otherwise

src/pytest_inmanta_lsm/orchestrator_container.py
```diff
@@ -44,8 +44,6 @@ def run_cmd(*, cmd: List[str], cwd: Path) -> Tuple[str, str]:
             env=env_vars,
         )
     except FileNotFoundError as e:
-        if e.filename == "docker-compose":
-            raise FileNotFoundError("The `docker-compose` command is not found. You need it to have a local orchestrator.")
```
edvgui, line 48: If this goes away, then catching the exception doesn't make sense anymore

jptrindade: Do you think it still makes sense keeping this check here? Is there another use for run_cmd besides the ones that I changed?

PR #460, opened 2024-11-19
Support both versions of disk layout file

jptrindade:
# Description

Related to https://github.com/inmanta/inmanta-core/pull/8376

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -465,7 +465,10 @@ class RemoteOrchestrator:
         This path depends on the on-disk layout of the remote orchestrator.
         """
         if self._remote_project_path is None:
-            cmd = "if test -f /var/lib/inmanta/.inmanta_use_new_disk_layout ; then echo True ; fi"
+            cmd = (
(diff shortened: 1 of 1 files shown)


src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -465,7 +465,10 @@ def remote_project_path(self) -> pathlib.Path:
         This path depends on the on-disk layout of the remote orchestrator.
         """
         if self._remote_project_path is None:
-            cmd = "if test -f /var/lib/inmanta/.inmanta_use_new_disk_layout ; then echo True ; fi"
+            cmd = (
+                "if test -f /var/lib/inmanta/.inmanta_use_new_disk_layout ||"
+                " -f /var/lib/inmanta/.inmanta_disk_layout_version; then echo True ; fi"
```
Hugo-Inmanta, line 470: Should we bother reading the actual version in the file and checking its value already?

jptrindade: I don't think it does much at the moment. We wouldn't do anything with that value. @wouterdb what do you think?

wouterdb (approved):

PR #462, opened 2024-11-22
Fix logic of inmanta disk layout version

jptrindade:
# Description

Used os.path to figure out if the file exists instead of bash

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -9,7 +9,6 @@
 import dataclasses
 import functools
 import logging
-import os.path
 import pathlib
 import shlex
 import subprocess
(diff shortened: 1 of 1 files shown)


src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -465,11 +466,10 @@ def remote_project_path(self) -> pathlib.Path:
         This path depends on the on-disk layout of the remote orchestrator.
         """
         if self._remote_project_path is None:
-            cmd = (
-                "if test -f /var/lib/inmanta/.inmanta_use_new_disk_layout ||"
-                " -f /var/lib/inmanta/.inmanta_disk_layout_version; then echo True ; fi"
+            base_directory = "/var/lib/inmanta/"
+            use_new_disk_layout: bool = os.path.isfile(f"{base_directory}.inmanta_use_new_disk_layout") or os.path.isfile(
+                f"{base_directory}.inmanta_disk_layout_version"
             )
```
wouterdb, line 472: This does it one wrong machine I think

wouterdb (changes requested):

wouterdb (approved):

PR #469, opened 2024-12-13
Issue/auto detect postgres version

edvgui:
# Description

If no postgres version is specified, resolve it from product compatibility info (instead of using hardcoded default one)

Part of https://github.com/inmanta/inmanta-core/issues/8464

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/orchestrator_container.py
+++ b/src/pytest_inmanta_lsm/orchestrator_container.py
@@ -197,7 +197,7 @@ class OrchestratorContainer:
         """
         self.compose_file = compose_file
         self.orchestrator_image = orchestrator_image
-        self.postgres_version = (
+        self.postgres_version = str(
             postgres_version
             or get_product_compatibility(get_image_version(self.orchestrator_image))["system_requirements"]["postgres_version"]
         )
--- a/src/pytest_inmanta_lsm/orchestrator_container.py
+++ b/src/pytest_inmanta_lsm/orchestrator_container.py
@@ -28,13 +28,40 @@ LOGGER = logging.getLogger(__name__)
 DOCKER_COMPOSE_COMMAND = None
 
 
+def run_cmd(*, cmd: List[str], cwd: Path) -> Tuple[str, str]:
+    """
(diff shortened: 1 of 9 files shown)


src/pytest_inmanta_lsm/orchestrator_container.py
```diff
@@ -19,13 +20,91 @@
...
+    )
+    if matched is None:
+        raise ValueError(f"Unsupported image format: {image}")
+
+    tag = matched.group("tag")
+    v = matched.group("version")
+
+    if tag in ["dev", "dev-ng"]:
+        # The is the latest dev version of the product
+        return version.Version("8.dev")
```
edvgui, line 48: Any suggestion on how to keep this pointing to the next major of the product?

Hugo-Inmanta: Disclaimer: very hackish:

```
import requests


def get_latest_released_iso() -> int:
    """
    Calls "https://docs.inmanta.com/inmanta-service-orchestrator/latest/" url
    and relies on the fact that there's a redirection (?) to
    "https://docs.inmanta.com/inmanta-service-orchestrator/7/index.html"
    with the proper latest released iso version in there.
    """
    url = "https://docs.inmanta.com/inmanta-service-orchestrator/latest/"
    resp = requests.get(url)

    resp.raise_for_status()
    return int(resp.url.removeprefix("https://docs.inmanta.com/inmanta-service-orchestrator/").removesuffix("/index.html"))


latest_dev_iso = get_latest_released_iso() + 1

print(latest_dev_iso)
```

Hugo-Inmanta (approved):
LGTM as far as I can see, added a suggestion to retrieve the latest master iso version but it's rather hackish. There's got to be a cleaner way

src/pytest_inmanta_lsm/orchestrator_container.py
edvgui: I had something like that in mind but didn't dare doing it.  But if I am not the only one thinking about it it might be worth it :p

wouterdb (approved):

PR #482, opened 2025-02-28
Don't reload config

edvgui:
# Description

The Config.set(...) call get_instance() which is a non-destructive way of loading the config, this should make it easier to use the rmeote orchestrator with other config-modifying fixtures

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -286,7 +286,6 @@ class RemoteOrchestrator:
         """
         Setup the config required to make it possible for the client to reach the orchestrator.
         """
-        inmanta_config.Config.load_config()
         inmanta_config.Config.set("config", "environment", str(self.environment))
 
         for section in ["compiler_rest_transport", "client_rest_transport"]:


src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -286,7 +286,6 @@ def setup_config(self) -> None:
         """
         Setup the config required to make it possible for the client to reach the orchestrator.
         """
-        inmanta_config.Config.load_config()
```
wouterdb, line 289: I think this will fail if the project fixture hasn't been loaded at all, because there is no config loaded at all. 
So you need to check here if it has been loaded...

wouterdb (changes requested):

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: I don't think so, because of 
- https://github.com/inmanta/inmanta-core/blob/d25205bdd49016596ad7653597a2cc99a8ed3992/src/inmanta/config.py#L230
- https://github.com/inmanta/inmanta-core/blob/d25205bdd49016596ad7653597a2cc99a8ed3992/src/inmanta/config.py#L162
- https://github.com/inmanta/inmanta-core/blob/d25205bdd49016596ad7653597a2cc99a8ed3992/src/inmanta/config.py#L154

wouterdb (approved):
I see

PR #484, opened 2025-03-02
Fix wait for

wouterdb:
# Description

Fix wait_until_deployment_finishes for iso8

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -951,7 +951,7 @@ class RemoteOrchestrator:
         else:
             self.wait_for_released(version)
 
-            def get_key_numbers():
+            def get_deployment_progress():
                 result = self.client.resource_list(self.environment, deploy_summary=True, limit=1)
                 assert result.code == 200
                 summary = result.result["metadata"]["deploy_summary"]
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -886,6 +886,21 @@ class RemoteOrchestrator:
             cwd=str(self.remote_project_path),
         )
 
+    def wait_for_released(self, version: int | None = None):
+        """
+        Version None means latest
+        """
+        retry_limited(functools.partial(self.is_released, version), timeout=3)
+
+    def is_released(self, version: int | None = None) -> bool:
+        """Version None means latest"""
+        versions = self.client.list_versions(tid=self.environment)
+        assert versions.code == 200
+        if version is None:
+            return versions.result["versions"][0]["released"]
+        lookup = {v["version"]: v["released"] for v in versions.result["versions"]}
+        return lookup[version]
+
     def wait_until_deployment_finishes(
         self,
         version: int,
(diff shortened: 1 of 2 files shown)


src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -899,24 +914,73 @@ def wait_until_deployment_finishes(
...
+                # {'by_state': {'available': 3, 'cancelled': 0, 'deployed': 12, 'deploying': 0, 'failed': 0, 'skipped': 0,
+                #               'skipped_for_undefined': 0, 'unavailable': 0, 'undefined': 0}, 'total': 15}
+                return (
+                    (
+                        summary["by_state"]["deployed"]
+                        + summary["by_state"]["failed"]
+                        + summary["by_state"]["skipped"]
+                        + summary["by_state"]["skipped_for_undefined"]
+                        + summary["by_state"]["unavailable"]
+                        + summary["by_state"]["undefined"]
```
edvgui, line 958: What about `cancelled`? Do we want to wait until the timeout to porentially see it? (not familiar with the meaning of that state).  I was wondering if we shouldn't go for `total - deploying - available` (feels like a more stable set of states as the one containing all the possible outcomes of a deployment)?

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -886,6 +886,21 @@ def export_service_entities(self) -> None:
             cwd=str(self.remote_project_path),
         )
 
+    def wait_for_released(self, version: int | None = None) -> None:
+        """
+        Version None means latest
+        """
```
edvgui, line 892:
```suggestion
        """
        Wait for a given version to be released by the orchestrator.
        :param version: The version to wait for, or None to wait for the latest.
        """
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -886,6 +886,21 @@ def export_service_entities(self) -> None:
...
+        retry_limited(functools.partial(self.is_released, version), timeout=3)
+
+    def is_released(self, version: int | None = None) -> bool:
+        """Version None means latest"""
```
edvgui, line 896:
```suggestion
        """
        Verify if a given version has already been released by the orchestrator.
        :param version: The version to check, or None to verify the latest version.
        """
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -886,6 +886,21 @@ def export_service_entities(self) -> None:
...
+        versions = self.client.list_versions(tid=self.environment)
+        assert versions.code == 200
```
edvgui, line 898:
```suggestion
        assert versions.code == 200, str(versions.result)
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -899,24 +914,73 @@ def wait_until_deployment_finishes(
...
-        def is_deployment_finished() -> bool:
-            response = self.client.get_version(self.environment, version)
-            assert response.result is not None
-            LOGGER.info(
-                "Deployed %s of %s resources",
-                response.result["model"]["done"],
-                response.result["model"]["total"],
-            )
-            return response.result["model"]["total"] - response.result["model"]["done"] <= 0
+        # Determin api version
```
edvgui, line 917:
```suggestion
        # Determine api version.  The resource engine of the orchestrator has evolved.  Checking the
        # layout of the api response allows us to know whether we are dealing with an old orchestrator
        # or the new resource scheduler.
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -899,24 +914,73 @@ def wait_until_deployment_finishes(
...
+            assert result.result is not None
+            for resource in result.result["resources"]:
+                LOGGER.info(f"Resource Status:\n{resource['status']}\n{pformat(resource, width=140)}\n")
+                assert (
+                    resource["status"] == desired_state
+                ), f"Resource status do not match the desired state, got {resource['status']} (expected {desired_state})"
+        else:
+            self.wait_for_released(version)
+
+            def get_key_numbers():
```
edvgui, line 945:
```suggestion
            def get_deployment_progress():
```
How about this?

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -899,24 +914,73 @@ def wait_until_deployment_finishes(
...
+                result = self.client.resource_list(self.environment, deploy_summary=True, limit=1)
+                assert result.code == 200
```
edvgui, line 947:
```suggestion
                assert result.code == 200, str(result.result)
```

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -899,24 +914,73 @@ def wait_until_deployment_finishes(
...
+                    total,
+                )
+                return total - done <= 0
+
+            retry_limited(is_deployment_finished, timeout)
+            result = self.client.get_version(self.environment, version)
+
+            assert result.result is not None
+            # We only check first 1000
+            result = self.client.resource_list(self.environment, limit=1000)
```
edvgui, line 978: I think we should use paging here to get all resources, or at the very least, explain why we only check 1000 (if there is a reason).  We rarely have more than 1000 resources in testing, but for scaling test (where we might actually use this method) we do

edvgui (changes requested):
Couples of comments, mostly cosmetic, some (less certain ones) related to my limited knowledge of the new resource scheduler

src/pytest_inmanta_lsm/remote_orchestrator.py
wouterdb: this is what we use on all the core tests, ... 

i.e. it is all non-transient states (everything has reached an endpoint )

src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -899,24 +915,73 @@ def wait_until_deployment_finishes(
...
+                        + summary["by_state"]["skipped_for_undefined"]
+                        + summary["by_state"]["unavailable"]
+                        + summary["by_state"]["undefined"]
+                    ),
+                    summary["by_state"]["failed"],
+                    summary["total"],
+                )
+
+            def is_deployment_finished() -> bool:
+                done, failed, total = get_key_numbers()
```
wouterdb, line 966:
```suggestion
                done, failed, total = get_deployment_progress()
```

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: Can we maybe reference this is in some way?

src/pytest_inmanta_lsm/remote_orchestrator.py
wouterdb: I agree, but I don't have the time to implement it. Could you add it?

edvgui: Paging is super annoying with the Client object, I created a ticket to improve support in core: https://github.com/inmanta/inmanta-core/issues/8801

In the meantime, I went for an alternative solution, we raise an exception after the first mismatching resource, so we can actually simply query the first mismatching resource from the api

src/pytest_inmanta_lsm/remote_orchestrator.py
edvgui: I think that `inmanta.const.DONE_STATES` is what we want, but it is not part of the stable api, can it be added?

edvgui (approved):

PR #488, opened 2025-03-24
Revert breaking change to service_entities

jptrindade:
# Description

`self.service_entities` is now `dict[str, ServiceEntity]` again and stores the default version of the service entity.
`get_service_entity` fetches from the appropriate diff given the provided version.
Also adds some version context to functions that were missing it.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -734,16 +734,22 @@ class LsmProject:
                 "Please call self.export_service_entities."
             )
 
-        key, entity_dict = (
-            ((service_entity_name, version), self.service_entity_versions)
-            if version is not None
-            else (service_entity_name, self.service_entities)
-        )
+        if service_entity_name not in self.service_entities:
+            raise LookupError(
+                f"Unknown service entity {service_entity_name}. Known services are: {list(self.service_entities.keys())}."
+            )
 
-        if key not in entity_dict:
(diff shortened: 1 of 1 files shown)


src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -744,18 +728,22 @@ def get_service_entity(self, service_entity_name: str, version: int | None = Non
...
+            ((service_entity_name, version), self.service_entity_versions)
+            if version is not None
+            else (service_entity_name, self.service_entities)
+        )
+
+        if key not in entity_dict:
+            raise LookupError(f"Unknown service entity {key}.  Known services are: {list(entity_dict.keys())}.")
 
-        return self.service_entities[(service_entity_name, version)]
+        return entity_dict[key]  # type: ignore[index]
```
edvgui, line 746: What does that one complain about?

edvgui: Why not update to code to avoid the type confusion?

```
if service_entity_name not in self.service_entities:
    # Service doesn't exist
    raise LookupError(...)

# If we don't want a specific version, take the default one
version = version or self.service_entities[service_entity_name].version

if (service_entity_name, version) not in self.service_entity_version:
    # Version doesn't exist
    raise LookupError(...)

return self.service_entity_version[(service_entity_name, version)]
```

jptrindade: wanted to "merge" the 2 lookup errors into one. But I guess like this is better and more readable

edvgui (approved):

PR #494, opened 2025-04-16
Rework LsmProject.compile interface

edvgui:
# Description

- Introduce two more explicits helpers: validating_compile and exporting_compile
- Update the old compile method to use the new ones and insert warning when mis-used
- Update auto transfer logic to also include exporting compiles

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -1016,9 +1016,7 @@ class LsmProject:
         if not self.partial_compile:
             # Specific services are requested for the compile but partial compile is not
             # supported, this is a user error
-            raise ValueError(
-                "Partial compile is not supported but a partial compile was attempted anyway"
-            )
+            raise ValueError("Partial compile is not supported but a partial compile was attempted anyway")
 
         # Verify that each service exists in the catalog
         for srv in service_ids:
@@ -1062,9 +1060,7 @@ class LsmProject:
         match (service_id, validation):
             case (None, True):
                 # Stay backward compatible, even though this is confusing
-                warnings.warn(
-                    "Validating compile requested without any service to validate.  Assuming full exporting compile."
-                )
+                warnings.warn("Validating compile requested without any service to validate.  Assuming full exporting compile.")
                 return self.exporting_compile(model=model)
             case (None, False):
                 # Normal full compile
@@ -1077,16 +1073,12 @@ class LsmProject:
                 return self.exporting_compile([service_id], model=model)
             case (collections.abc.Sequence(), True):
                 # Validation of multiple services, this is not allowed
-                raise ValueError(
-                    "Validating compile can not be done for multiple instances at once"
-                )
+                raise ValueError("Validating compile can not be done for multiple instances at once")
             case (collections.abc.Sequence(), False):
                 # Partial compile
                 return self.exporting_compile(service_id, model=model)
             case _:
-                raise ValueError(
-                    f"Unexpected input value combination: service_id={service_id}, validation={validation}"
-                )
+                raise ValueError(f"Unexpected input value combination: service_id={service_id}, validation={validation}")
 
     def post_partial_compile_validation(
         self,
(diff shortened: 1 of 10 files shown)


edvgui:
Something doesn't work with connect, I need to dig

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -926,81 +936,99 @@ def add_service(
...
-                warnings.warn("Partial compile is enabled but it is not supported, it will be ignored.")
+        if service_ids is None:
+            # Triggering a full compile
+            LOGGER.debug("Triggering full compile")
+            self.project.compile(model, no_dedent=False)
+            return
+
+        if not self.partial_compile:
+            # Specific services are requested for the compile but partial compile is not
+            # supported, this is a user error
```
wouterdb, line 1018: What does not supported mean in this context?

i.e. if someone want to run a test in partial compile mode, what causes us to prevent that?

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -926,81 +936,99 @@ def add_service(
...
+        model = self._get_model(model)
 
-        try:
-            env[inmanta_lsm.const.ENV_PARTIAL_COMPILE] = str(self.partial_compile and bool(service_ids))
-        except AttributeError:
-            # This attribute only exists for iso5+, iso4 doesn't support partial compile.
-            # We then simply don't set the value.
-            if self.partial_compile:
-                warnings.warn("Partial compile is enabled but it is not supported, it will be ignored.")
+        if service_ids is None:
```
wouterdb, line 1010: You could also do a full compile with the service id set? (it can be considered a bit weird, but that may be the point of the test?)

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -1009,6 +1037,58 @@ def compile(
...
+                    )
+                    return self.exporting_compile(model=model)
+            case (collections.abc.Sequence(), True):
+                # Validation of multiple services, this is not allowed
+                raise ValueError("Validating compile can not be done for multiple instances at once")
+            case (collections.abc.Sequence(), False):
+                # Partial compile
+                return self.exporting_compile(service_id, model=model)
+            case _:
+                raise ValueError(f"Unexpected input value combination: service_id={service_id}, validation={validation}")
```
wouterdb, line 1090: Cool!

src/pytest_inmanta_lsm/lsm_project.py
edvgui: The issue I have with it is that doing this really doesn't mean something (it is an undefined scenario) so if someone requests this, they probably don't know what they are requesting (or if they do, they might not realize that the testing library doesn't consider this to be a use case)

src/pytest_inmanta_lsm/lsm_project.py
edvgui: Not supported whean that `LsmProject.partial_compile` is set to `False`, so a test can be run in partial compile mode, it just needs to be set on the class

wouterdb: how is it set?

edvgui: The default value will depend on `INMANTA_LSM_PARTIAL_COMPILE` env var: https://github.com/inmanta/pytest-inmanta-lsm/blob/cb32891661e4cb93990ea444afe66d1eacb91bb7/src/pytest_inmanta_lsm/plugin.py#L110

But in a partial compile specific test, you will usually set it manually anyway: https://code.inmanta.com/solutions/modules/connect/-/blob/7b71db1e242243a12d2b9e151e1982198fbff752/tests/test_model/test_l2_connect.py#L73

wouterdb: I find this confusing, but I'm not the primary user. 
I would at least mention in the exception how to resolve the problem.

edvgui: I updated it, is it more clear?

wouterdb (approved):

PR #495, opened 2025-04-23
Issue/fix token auth

edvgui:
# Description

Fix support for orchestrators with authentication enabled:
- Setup client config before constructing clients
- Persist client token in local inmanta config file to sync it to the remote orchestrator

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/remote_orchestrator.py
+++ b/src/pytest_inmanta_lsm/remote_orchestrator.py
@@ -15,6 +15,7 @@ import subprocess
 import typing
 import urllib.parse
 import uuid
+import toml
 from pprint import pformat
 from uuid import UUID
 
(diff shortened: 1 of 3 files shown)


src/pytest_inmanta_lsm/remote_orchestrator.py
```diff
@@ -301,6 +302,23 @@ def setup_config(self) -> None:
...
+        }
+        if self.token:
+            raw_config["compiler_rest_transport"]["token"] = self.token
+            raw_config["client_rest_transport"]["token"] = self.token
+
+        # Persist environment and token info in the inmanta config file of the project
+        # to make sure it is sent to the remote orchestrator
+        project_path = pathlib.Path(self.local_project._path)
+        config_file_path = project_path / ".inmanta"
+        config_file_path.write_text(toml.dumps(raw_config))
```
wouterdb, line 320: Why/ when is this required? i.e. the server should set this up correctly for the compiler when the server runs the compiler?

edvgui: It failed when running the export service entities plugin we run during the sync of the project, I guess because this compile is not triggered by the orchestrator itself?

edvgui: We could try to trigger the export via the api instead of doing it manually, that might be a cleaner approach, but more work to implement, I would then defer it to another ticket

PR #519, opened 2025-08-19
Issue/improve service export testing

edvgui:
# Description

- Improve export_service_entities functionality in LsmProject and RemoteOrchestrator to mimic the lsm extension better.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/src/pytest_inmanta_lsm/lsm_project.py
+++ b/src/pytest_inmanta_lsm/lsm_project.py
@@ -35,6 +35,12 @@ INMANTA_LSM_MODULE_NOT_LOADED = (
     "    - If you are using v2 modules: make sure the inmanta-module-lsm is installed in your venv."
 )
 
+try:
+    from inmanta_lsm.const import ENV_NO_INSTANCES
+except ImportError:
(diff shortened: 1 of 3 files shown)


src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -35,6 +35,12 @@
     "    - If you are using v2 modules: make sure the inmanta-module-lsm is installed in your venv."
 )
 
+try:
+    from inmanta_lsm.const import ENV_NO_INSTANCES
+except ImportError:
+    # Ensure backwards compatibility with older versions of the inmanta-lsm extensions.
+    ENV_NO_INSTANCES = "lsm_no_instances"
```
cukiernick, line 42: Do we need this? The const was added 18 months ago (inmanta-lsm 5.0.0)

cukiernick (approved):

src/pytest_inmanta_lsm/lsm_project.py
cukiernick: now I see ISO 7.4.x is still on inmanta-lsm 4

Issue #531, opened 2025-10-15
fix race condition in wait_until_deployment_finishes


Hugo-Inmanta:
`wait_until_deployment_finishes` waits for a version to be released and then deployed. It should wait for the version to be scheduled (use `list_desired_state_versions` endpoint) in an additional step between the first two.

When the scheduling takes significantly longer than the releasing, this method incorrectly reports that the deployment is finished for this version.

PR #535, opened 2025-11-05
Issue/track release time follow up

Hugo-Inmanta:
# Description

Follow-up / part of https://github.com/inmanta/pytest-inmanta-lsm/issues/531

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.15.0
+current_version = 4.0.0
 tag = False
 commit = False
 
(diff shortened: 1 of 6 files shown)


sanderr (approved):

src/pytest_inmanta_lsm/__init__.py
```diff
@@ -13,20 +13,21 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def retry_limited(fun: Callable[[], bool], timeout: int) -> None:
+def retry_limited(fun: Callable[[], bool], *, timeout: int, retry_interval: float = 1.0) -> None:
```
Hugo-Inmanta: This is probably better for absolute backwards compatibility :
```suggestion
def retry_limited(fun: Callable[[], bool], timeout: int, *, retry_interval: float = 1.0) -> None:
```

PR #569, opened 2026-07-06
Add RemoteOrder to drive services through the order API

edvgui:
# Description

Add sync and async `RemoteOrder` classes, mirroring `RemoteServiceInstance`, that wrap the orchestrator order API (`lsm_order_create` / `lsm_order_get`) using the typed objects from `inmanta_lsm.order.model`. They let a test create/update/delete one or more service instances in a single atomic order and wait for the order to reach a terminal state.

- `remote_order_async.RemoteOrder`: async class with `create`, `get`, `wait_for_state`, and the `RemoteOrderError` / `BadOrderStateError` / `OrderStateTimeoutError` exceptions.
- `remote_order.RemoteOrder`: sync wrapper, same proxy pattern as `remote_service_instance`.
- Generated `.pyi` stub + `Makefile` `stub` target updated.
- Quickstart tests extended (`test_order_full_cycle`, `test_order_sync`) and README documented.

closes *no ticket*

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation

### DIFF
--- a/examples/quickstart/tests/test_quickstart.py
+++ b/examples/quickstart/tests/test_quickstart.py
@@ -12,6 +12,7 @@ import time
 import uuid
 
 import inmanta_lsm.model
+import inmanta_lsm.order.model
 import pytest
 from pytest_inmanta import plugin
 
@@ -19,6 +20,8 @@ import pytest_inmanta_lsm.lsm_project
 from pytest_inmanta_lsm import (
     load_generator,
     remote_orchestrator,
+    remote_order,
+    remote_order_async,
     remote_service_instance,
     remote_service_instance_async,
     util,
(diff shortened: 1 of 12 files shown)


examples/quickstart/tests/test_quickstart.py
```diff
@@ -259,6 +262,148 @@ def test_transient_state(project: plugin.Project, remote_orchestrator: remote_or
...
+                },
+            ),
+        ],
+        wait_for_state=inmanta_lsm.order.model.OrderState.success,
+        timeout=60,
+    )
+
+    # Assert that the order reached the success state and manages our instance
+    assert service_order.status.state == inmanta_lsm.order.model.OrderState.success
+    assert {item.instance_id for item in service_order.service_order_items} == {instance_id}
```
DariushMoh, line 465: so we can get all the created services out of the order returned ? I think that would be handy to test if allocated values are correctly set for example

DariushMoh: or to delete only one service out of the order.

edvgui: Actually, you need to provide the service id as input for the order api, so you already have everything you need to be able to pull it after the order is complete.  But you can indeed also find it back in the response of the order.

DariushMoh (approved):

src/pytest_inmanta_lsm/remote_order_async.py
```diff
@@ -0,0 +1,396 @@
...
+        :raises OrderStateTimeoutError: If the timeout is reached while waiting for the desired state
+        """
+        if timeout is None:
+            timeout = self.DEFAULT_TIMEOUT
+
+        if bad_states is None:
+            bad_states = [state for state in self.ALL_BAD_STATES if state != target_state]
+
+        # Save the start time to know when we should trigger a timeout error
+        start = time.time()
```
arnaudsjs: In general it's recommended to rely on `loop.time()` or `time.monotonic()` for this. These are monotonic clock so they are not affected by changes to the system clock and can never move backwards.

arnaudsjs (approved):

src/pytest_inmanta_lsm/remote_order_async.py
edvgui: Oh, I didn't know about that, thanks!

PR #572, opened 2026-07-16
Add opt-in compiler reuse for the lsm_project fixture

edvgui:
> Created by Claude on behalf of Guillaume.

# Description

A single `lsm_project`-based test compiles the *same* model many times, only varying the LSM service data (attribute sets, current state, ...) injected through environment variables and read during the **execution** phase. The parse and type-definition phases produce an identical result on every compile, yet the stock harness rebuilds them each time — creating a fresh `module.Project`, re-reading the whole import closure and de-serialising every module AST from the parser cache. For a large model that AST de-serialisation dominates the test wall-clock time.

This adds an **opt-in** `CompileCache` (`compile_cache.py`) that:
- parses and type-checks a given model text once (first/"cold" compile), keeping the loaded project + typed program in memory;
- for every subsequent ("warm") compile of that same model, resets only the per-compile *execution* state left on the shared typed graph and re-runs the execution phase.

The exported desired state is identical to a normal compile.

**Measured on a large downstream module test** (12 compiles): wall-clock dropped from ~127s to ~35s (~3.6x), with byte-for-byte identical exported resources after accounting for the `requires` set-ordering that is already non-deterministic between two stock compiles.

**Enable it** with `--lsm-reuse-compiler` / `INMANTA_LSM_REUSE_COMPILER`, or by overriding the `lsm_reuse_compiler` fixture. Disabled by default.

**Safety**: the cache reaches into inmanta-core / pytest-inmanta internals (compiler entry points, the scheduler, and a few AST node fields). It is guarded by a capability check that fails loudly (`CompilerReuseUnavailableError`) rather than silently falling back when those internals are not shaped as expected, so a core upgrade can never silently corrupt compile output. The patches are installed via the test's `monkeypatch` fixture, so they are auto-undone per test and never leak.

closes n/a

# Self Check:

- [x] Changelog entry
- [x] ~~End user documentation~~ (option is self-documenting via `--help`; behaviour documented in `compile_cache` module docstring and CHANGELOG)

### DIFF
--- a/src/pytest_inmanta_lsm/compile_cache.py
+++ b/src/pytest_inmanta_lsm/compile_cache.py
@@ -19,8 +19,11 @@ time of the test.
 
 This module makes that work happen once.  The first compile of a given model
 text is a normal compile; afterwards the loaded project, its typed AST and the
-resulting types are kept in memory and reused.  Each subsequent ("warm") compile
-skips parse and ``define_types`` and only:
+resulting types are kept in memory and reused.  Because the parsed AST never
+leaves memory, the on-disk parser cache is redundant while the feature is
+active, so it is turned off for the test: that removes the (otherwise dominant)
+cost of writing every module's ``.cfc`` during the single cold compile.  Each
+subsequent ("warm") compile skips parse and ``define_types`` and only:
 
     * resets the per-compile *execution* state that the previous compile left on
       the shared typed graph, and
@@ -146,6 +149,8 @@ class CompileCache:
             (compiler_mod, "ProjectLoader"),
             (compiler_mod.ProjectLoader, "_reset_module_state"),
             (compiler_config, "track_dataflow"),
+            (compiler_config, "feature_compiler_cache"),
+            (compiler_config.feature_compiler_cache, "get"),
             (scheduler_mod.Scheduler, "define_types"),
             (scheduler_mod.Scheduler, "run"),
             (PytestInmantaProject, "_create_project_and_load"),
@@ -165,8 +170,17 @@ class CompileCache:
         self._check_capabilities()
 
         import inmanta.compiler as compiler_mod
+        from inmanta.compiler import config as compiler_config
         from inmanta.execute import scheduler as scheduler_mod
         from pytest_inmanta.plugin import Project as PytestInmantaProject
+
+        # The on-disk parser cache is redundant here: the in-memory typed program
+        # is reused across every compile of the test, so warm compiles never parse
+        # and thus never read the cache.  Leaving it enabled only pays the cost of
+        # writing every module's .cfc during the single cold compile (dominant for
+        # a large import closure) for a benefit that is never collected.  Disable
+        # it for the duration of the test; monkeypatch reverts it afterwards.
+        monkeypatch.setattr(compiler_config.feature_compiler_cache, "get", lambda: False)
 
         orig_create = PytestInmantaProject._create_project_and_load
         orig_define_types = scheduler_mod.Scheduler.define_types
--- a/src/pytest_inmanta_lsm/compile_cache.py
+++ b/src/pytest_inmanta_lsm/compile_cache.py
@@ -1,0 +1,317 @@
+"""
+Pytest Inmanta LSM
+
+:copyright: 2026 Inmanta
+:contact: code@inmanta.com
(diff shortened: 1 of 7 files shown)


src/pytest_inmanta_lsm/compile_cache.py
```diff
@@ -0,0 +1,317 @@
...
+"""
+
+from __future__ import annotations
+
+import logging
+import typing
+
+import pytest
+
+if typing.TYPE_CHECKING:
```
sanderr, line 57: Nice. Good call in this case imo

src/pytest_inmanta_lsm/compile_cache.py
```diff
@@ -0,0 +1,317 @@
...
+
+        for entity in state.types.values():
+            if isinstance(entity, Entity):
+                entity._index.clear()
+                entity._instance_list.clear()
+                entity.index_queue.clear()
+
+        assert state.root_ns is not None  # guaranteed once state.compiled is True
+        for namespace in self._all_namespaces(state.root_ns):
+            namespace.scope = None
```
sanderr, line 326: This part seems to reset state on type-level information (something I'd argue is probably not a great idea in the first place, but it is what it is). Resetting that state would make sense to me if we in fact kept all the type information between compiles. But I see that above you call `sched.run`, which re-evaluates the types.

It's not wrong per se (I'd have to dive deeper to judge that), but nags at me as a small inconsistency that could mean you're missing a part of the reset (i.e. how comes `state.types` still has the entities if we're redefining them anyway?), or an opportunity to keep the typing information warm (not the index information, but perhaps the rest of entity state, and the namespace scope).

src/pytest_inmanta_lsm/compile_cache.py
```diff
@@ -0,0 +1,317 @@
...
+
+        for statement in state.statements:
+            if isinstance(statement, DefineRelation):
+                refreshed: list[tuple[ResultVariable, ExpressionStatement]] = [
+                    (ResultVariable(), exp) for (_, exp) in statement.annotation_expression
+                ]
+                statement.annotation_expression = refreshed
```
sanderr, line 333: I wanted to comment that we could safely keep this warm, considering that we have the exact same model. But I was surprised to see how dynamic these actually are. Could be an opportunity to make this a bit more static. If `lsm::all` and the likes could not impact this, we could safely keep it warm.

src/pytest_inmanta_lsm/compile_cache.py
edvgui: _Claude on behalf of Guillaume:_

Good catch on the apparent inconsistency, but the typing *is* kept warm here. In a warm compile `define_types` is short-circuited (the patched version returns early for the warm scheduler, see `install`/`_warm_schedulers`), so `sched.run` only runs the execution phase — it does not re-evaluate the types. That is why `state.types` legitimately keeps the entities across compiles: `_reset_execution_state` only has to clear the *execution* state those (reused) type objects accumulated (constructed instances, index entries, namespace scope). I added a comment in `_warm_compile` to make this explicit.

src/pytest_inmanta_lsm/compile_cache.py
edvgui: _Claude on behalf of Guillaume:_

Agreed, I was surprised too. These annotation `ResultVariable`s are created at parse time on the AST node itself but only *set* during execution, so with the current core internals they cannot safely be kept warm — hence the defensive reset. Making the immutable AST truly reusable (separating it from per-compile execution state) is exactly the kind of core-side change captured in inmanta/inmanta-core#10597. Once that lands this reset can go away.

src/pytest_inmanta_lsm/compile_cache.py
edvgui: _Claude on behalf of Guillaume:_

Thanks! Keeping these imports under `TYPE_CHECKING` was deliberate: none of them are needed at runtime (they only annotate the cache internals), and pulling the compiler internals into the import graph eagerly would both slow down plugin import and defeat the whole point of the capability check, which is meant to be the single, explicit place where we probe for those internals. So the module stays cheap to import and degrades loudly rather than at import time. Appreciate the call-out.

PR #581, opened 2026-08-05
Support owned services in post_partial_compile_validation

edvgui:
# Description

`LsmProject.post_partial_compile_validation` assumed that the service being validated owns its
own resource set.  Services which are owned by another service (via the `owner` and
`relation_to_owner` attributes of their service entity binding) don't get one: they push their
owned resources up to the root of their ownership tree, which is the only service emitting a
`std::ResourceSet`.  Validating such a service always failed, so projects had to re-implement
the method themselves.

The validation now resolves the root of the ownership tree of the service it is given, and
validates the resource set of that root, which holds the resources of every service in the tree.
A resource set is expected iff at least one service of the tree is in an exporting state, so
deleting an owned service no longer wrongly expects its owner's resource set to disappear.  The
final full-compile check expects one resource set per ownership tree instead of one per service.

Three helpers are exposed for users who still need their own validation logic:
`LsmProject.get_owner`, `LsmProject.get_owner_root` and `LsmProject.exporting_resource_sets`.

Tested in `examples/test-multi-version`, which already models a parent service owning a child
service: ownership resolution, validation of both the child and its owner against the shared
resource set, update of the owner while the child exists, and deletion of the child and then of
the owner.

closes #580

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included or an issue is created for end-user documentation (the method's docstring is the end user documentation)


src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -1130,15 +1185,22 @@ def post_partial_compile_validation(
...
-        any other service.
+        Services which are owned by another service (as defined by the `owner` and
+        `relation_to_owner` attributes of their service entity binding) don't have a resource set
+        of their own: they contribute their resources to the resource set of the service at the
+        root of their ownership tree.  Validating such a service therefore validates the resource
+        set of that root, which contains the resources of every service in the tree.  The owned
+        resource patterns are then expected to match the resources of the full tree.
+
+        This method only works with services which don't need any service outside of their own
+        ownership tree to be present in the partial compile with them.
```
edvgui: Why? This should also be fine, the module developer just has to extend the partial selection logic to also pull in these additional services.

src/pytest_inmanta_lsm/lsm_project.py
```diff
@@ -1147,16 +1209,21 @@ def post_partial_compile_validation(
...
 
         resource_sets = get_resource_sets(self.project)
-        if not service.deleted:
-            # Check that the only resource set emitted is the one of this service
-            assert resource_sets.keys() == {str(service_id)}
-            _, owned_resources = resource_sets.popitem()
+        if exports_resources:
+            # Check that the only resource set emitted is the one of this ownership tree
+            assert resource_sets.keys() == {str(root_id)}
+            owned_resources = resource_sets[str(root_id)]
```
edvgui: This is suspicious, it looks to me that if a service has an interservice relationship to another service, in a different resource set (ie l2Connect and UserNetworkInterface in connect module), this validation would fail.  But we currently have tests in the connect module which seem to work, relying on this function.  How come?

src/pytest_inmanta_lsm/lsm_project.py
edvgui: *(Claude, on behalf of Guillaume)*

Agreed, that limitation was inherited from the old implementation and there was no good reason to keep it. Removed in 587f922.

The validation now accepts a partial compile emitting more than one resource set, but only for services that are declared explicitly, through a new `additional_services` argument:

```python
lsm_project.post_partial_compile_validation(
    uni_service.id,
    shared_resources,
    owned_resources,
    additional_services=[ne_service.id],
)
```

Expected resource sets are the ones of the ownership trees of the validated service and of the additional services (intersected with the trees that actually have an exporting service). I kept it explicit rather than just accepting any set belonging to some exporting service, because the point of the check is to catch a partial compile whose scope is broader than intended, and a plain subset check would silently let that through.

Resources that are part of any of the emitted sets are matched against the owned patterns; the shared patterns keep covering exactly the resources that are in no resource set.

src/pytest_inmanta_lsm/lsm_project.py
edvgui: *(Claude, on behalf of Guillaume)*

Good catch, and the answer is that the connect module does not use this function. It has its own copy in `tests/helpers.py:252`, which differs on exactly this line:

```python
# connect, tests/helpers.py
assert str(service.id) in resource_sets
owned_resources = list(itertools.chain(*resource_sets.values()))
```

Membership instead of equality, and the union of every emitted set as the owned resources. That is why `test_ne_linked_services.py:129` passes while partial-compiling a UNI that pulls in the NE service: two resource sets are emitted and the local copy tolerates it. The version in this repo never supported it, on master no more than in the first commit of this PR.

So the suspicion was right, and 587f922 fixes it. Multiple resource sets are now supported, with the extra ones declared through `additional_services` (see the other thread for why they are declared rather than just tolerated). Resource classification matches the connect helper: in a resource set means owned, in no resource set means shared.

That should let connect drop its fork. The differences left with the local copy are that this one is stricter about undeclared resource sets, and that it still asserts on a resource matched by both the shared and the owned patterns, which the connect copy has commented out for netbox vlans. Happy to add an escape hatch for that too if you want to do the migration.

src/pytest_inmanta_lsm/lsm_project.py
edvgui: *(Claude, on behalf of Guillaume)*

Switched to two modes in b4cb65b, the strict check is now opt-in:

- `additional_services=None` (default): only the presence of the resource set of the validated service's ownership tree is checked, any other resource set the partial compile emitted is tolerated. This is what the connect helper does today.
- `additional_services=[...]`, including `[]`: the emitted resource sets must be exactly the ones of the validated service and of the declared additional services.

Resource classification is the same in both modes: in a resource set means owned, in no resource set means shared.

examples/test-multi-version/tests/test_partial_compile.py
```diff
@@ -0,0 +1,232 @@
...
+    """
+    create_tree(lsm_project)
+
+    # The child doesn't emit a resource set of its own, its resources are part of the
+    # resource set of its owner
+    resource_sets = get_resource_sets(lsm_project.project)
+    assert resource_sets.keys() == {str(PARENT_ID)}
+    assert {str(resource) for resource in resource_sets[str(PARENT_ID)]} == {
+        f"lsm::LifecycleTransfer[lsm,instance_id={PARENT_ID}]",
+        f"lsm::LifecycleTransfer[lsm,instance_id={CHILD_ID}]",
```
edvgui: iso8 will have internal agent

examples/test-multi-version/tests/test_partial_compile.py
```diff
@@ -0,0 +1,232 @@
...
+    child.deleted = True
+    child.version += 1
+    lsm_project.exporting_compile([child.id])
+    lsm_project.post_partial_compile_validation(child.id, SHARED_RESOURCES, OWNED_RESOURCES)
+
+    # The child's resources are gone, but the owner's resource set is still there
+    resource_sets = get_resource_sets(lsm_project.project)
+    assert resource_sets.keys() == {str(PARENT_ID)}
+    assert {str(resource) for resource in resource_sets[str(PARENT_ID)]} == {
+        f"lsm::LifecycleTransfer[lsm,instance_id={PARENT_ID}]",
```
edvgui: iso8 will have internal agent

examples/test-multi-version/tests/test_partial_compile.py
edvgui: *(Claude, on behalf of Guillaume)*

Right, hardcoding `lsm` there would have broken on iso8. Fixed in 232e1ff by dropping the agent from the comparison entirely, rather than branching on the version: the assertions now compare the resource type and the identifying attribute value.

```python
def resource_set_content(lsm_project, resource_set) -> set[tuple[str, str]]:
    return {
        (resource.entity_type, resource.attribute_value)
        for resource in get_resource_sets(lsm_project.project)[str(resource_set)]
    }

assert resource_set_content(lsm_project, PARENT_ID) == {
    ("lsm::LifecycleTransfer", str(PARENT_ID)),
    ("lsm::LifecycleTransfer", str(CHILD_ID)),
    ("std::testing::NullResource", "parent"),
    ("std::testing::NullResource", "child"),
}
```

That keeps the part these two assertions exist for, which is that the child's resources land in its owner's resource set and leave it again when the child is deleted, and it no longer cares which agent the resources are exported to. The `OWNED_RESOURCES` patterns were already agent-agnostic (`\[.*,name=parent\]`), so nothing else in the file pins an agent name now.

examples/test-multi-version/tests/test_partial_compile.py
edvgui: *(Claude, on behalf of Guillaume)*

Right, hardcoding `lsm` there would have broken on iso8. Fixed in 232e1ff by dropping the agent from the comparison entirely, rather than branching on the version: the assertions now compare the resource type and the identifying attribute value.

```python
def resource_set_content(lsm_project, resource_set) -> set[tuple[str, str]]:
    return {
        (resource.entity_type, resource.attribute_value)
        for resource in get_resource_sets(lsm_project.project)[str(resource_set)]
    }

assert resource_set_content(lsm_project, PARENT_ID) == {
    ("lsm::LifecycleTransfer", str(PARENT_ID)),
    ("lsm::LifecycleTransfer", str(CHILD_ID)),
    ("std::testing::NullResource", "parent"),
    ("std::testing::NullResource", "child"),
}
```

That keeps the part these two assertions exist for, which is that the child's resources land in its owner's resource set and leave it again when the child is deleted, and it no longer cares which agent the resources are exported to. The `OWNED_RESOURCES` patterns were already agent-agnostic (`\[.*,name=parent\]`), so nothing else in the file pins an agent name now.

PR #588, opened 2026-08-14
improve feedback

### LINKED ISSUE #587: Improve order api diagnose for non-compliant resources
We might reach a fail state from a resource that is not compliant. 

The goal of this ticket is to add checks on the service for non-compliant resources that may have cause the transfer. And add the diagnose the diff of non-compliant resources

FelixGaudin:
# Description

* Short description here *

closes https://github.com/inmanta/pytest-inmanta-lsm/issues/587

# Self Check:

- [ ] Changelog entry


src/pytest_inmanta_lsm/remote_service_instance.pyi
```diff
@@ -19,6 +22,40 @@ def get_service_instance_from_log(log: model.ServiceInstanceLog) -> model.Servic
...
+    :param last_handler_run_at: When the handler of the resource has last been run.
+    :param attribute_diff: For a non-compliant resource, the deviation of each attribute
+        which doesn't have its desired value.
+    """
+
+    report_only: bool
+    compliance: str
+    last_handler_run: str
+    last_handler_run_at: datetime.datetime | None
+    attribute_diff: dict[str, AttributeStateChange] | None
```
edvgui: Don't we have typed object for the response of the api already defined in inmanta-core?

edvgui: Oh, or is this generated? Weird that it duplicates these

edvgui: Ah the real definition is in remote orch, still, I think it could be dropped in favor of the core types

FelixGaudin: Good catch

DariushMoh (approved):

src/pytest_inmanta_lsm/remote_service_instance.pyi
```diff
@@ -1,14 +1,18 @@
+import datetime
 import typing
 import uuid
 
+import pydantic
 from _typeshed import Incomplete
+from inmanta.data.model import AttributeStateChange as AttributeStateChange
```
edvgui:
```suggestion
from inmanta.data.model import AttributeStateChange
```

src/pytest_inmanta_lsm/remote_service_instance.pyi
edvgui: Only the first class was fixed, second one is still pending

FelixGaudin: done

src/pytest_inmanta_lsm/remote_service_instance.pyi
FelixGaudin: done

edvgui (approved):
Cool!

PR #590, opened 2026-08-21
Log hanging services on order timeout

### LINKED ISSUE #589: Add information about which service are hanging when we reach timout for order api
Before when deploying service per service we could know which service was guilty.
When we get an error with order api we don't have that information.

The goal of this ticket is to check which service(s) is/are hanging

FelixGaudin:
# Description

When getting a timeout with an order, show which services are hanging

closes https://github.com/inmanta/pytest-inmanta-lsm/issues/589

# Self Check:

- [ ] Changelog entry


DariushMoh (approved):

edvgui (approved):
Looks good.  Can you fix the title and description of the MR?

